Issue #338, opened 2017-11-28
Spelling mistakes have not influence on the result of  "deallacate prepare statement_name" 


zhunina:
```
mysql> deallacate prepare stmt
    -> ;
Query OK, 0 rows affected (0.00 sec)

mysql> execute stmt using @c;
ERROR 1243 (HY000): Unknown prepared statement handlerSTMT given to EXECUTE
```

ditdb:
已修正，来自mycat的bug

zhunina:
fixed

PR #343, opened 2017-11-29
fix incorrect key being ignored

ditdb:
Reason: 
when the first key is incorrect， the parser skips it and continues parseing. 
eg. 
shsh create table xxx(a int, b char(32))；

when the statement was parse, the shsk will be skiped, the actual result is executing：

create table xxx(a int, b char(32))；

This is wrong. The bug was brought in from the commit ff7f9160acff462fdc02bae9e6f2cf99c5fb1dc0 in mycat. But I don't find the reason and a statement that should be parsed by the way. Maybe the writer only get to the root of the matter. I just fix it.

Type:  fix

Influences：  
all statement.


yanhuqing666:
#338

PR #483, opened 2018-03-26
issue 436 union fix

sunsun314:
Reason:
  union fix
Type:
  Bug fix
Influences：
  Change the way to terminate a union query


joycezhou007:
add case in automation, union_function_shard.sql:

#!share_conn
select id from  test_shard union SELECT OLD_PASSWORD('mypass')
select id from  test_shard union SELECT SHA('abc')

Issue #345, opened 2017-11-29
Error when `view` contains a statement which like`select table1.name,table2.name as t_name`


FlyingMao:
steps：
```
mysql> CREATE TABLE a_test(`id` int(10) unsigned NOT NULL,`t_id` int(10) unsigned NOT NULL DEFAULT '0',`name` char(120) NOT NULL DEFAULT '',`pad` int(11) NOT NULL,PRIMARY KEY (`id`),KEY `k_1` (`t_id`));
Query OK, 0 rows affected (0.31 sec)

mysql> CREATE TABLE a_order(`id` int(10) unsigned NOT NULL,`o_id` int(10) unsigned NOT NULL DEFAULT '0',`name` char(120) NOT NULL DEFAULT '',`pad` int(11) NOT NULL,PRIMARY KEY (`id`),KEY `k_1` (`o_id`))
    -> ;
Query OK, 0 rows affected (0.83 sec)

mysql> insert into a_test values(1,1,'test中id为1',1),(2,2,'test_2',2),(3,3,'test中id为3',4),(4,4,'$test$4',3),(5,5,'test...5',1),(6,6,'test6',6);
Query OK, 6 rows affected (0.12 sec)

mysql> insert into a_order values(1,1,'order中id为1',1),(2,2,'test_2',2),(3,3,'order中id为3',3),(4,4,'$order$4',4),(5,5,'order...5',1);
Query OK, 5 rows affected (0.02 sec)

mysql> create view view_test as select a_test.id,a_test.name,a_test.pad,a_order.name as a_name from a_test inner join a_order;
ERROR 1999 (HY000): Duplicate column name 'name'
```
logs:
```
2017-11-29 16:38:20.416 DEBUG [$_NIOREACTOR-1-RW] (com.actiontech.dble.net.FrontendConnection.query(FrontendConnection.java:296)) - ServerConnection [id=1, schema=mytest, host=127.0.0.1, user=test,txIsolation=3, autocommit=true, schema=mytest] create view view_test as select a_test.id,a_test.name,a_test.pad,a_order.name as a_name from a_test inner join a_order
[...]

PR #578, opened 2018-06-27
#345 Duplicate column name

sunsun314:
Reason:
  BUG#345
Type:
  BUG
Influences：
  Add skip to default alias 'autoalias_scalar'


yanhuqing666:
reverted

Issue #888, opened 2018-12-12
dble manager has not a way to kill in-processing ddl but expect


irene-coming:
dble version:
5.6.29-dble-9.9.9.9-3eb26a6-20181210020901

yanhuqing666:
add it in show @@help #852

yanhuqing666:
doc need update @Baofengqi

yexiaoli88:
function already has been verified on dble version 5.6.23-dble-9.9.9.9-d52a65fafdd1cbf36151f7d51a37f08d055dcb84-20190219080310

Baofengqi:
doc updated

yexiaoli88:
verified on version 5.6.23-dble-9.9.9.9-c3445b55c3e57110427b5b0862a56411fd9f96b7-20190221102958

PR #987, opened 2019-02-14
modify tip of 'kill ddl' cmd

Baofengqi:
Reason:  
  issue #888
Type:  
       Issue  
Influences：  
  modify tip of 'kill ddl' cmd


sunsun314:
OK

Issue #788, opened 2018-10-30
Some syntax, druid parser does not recognize


FlyingMao:
* **dble version：**  
5.6.29-dble-9.9.9.9-359645c-20181029082136
* **preconditions ：**  
no
* **configs：** 

**schema.xml**  

```
<?xml version="1.0"?>
<!DOCTYPE dble:schema SYSTEM "schema.dtd">
<dble:schema xmlns:dble="http://dble.cloud/">
	<schema name="mytest" sqlMaxLimit="100" dataNode="dn5">
        <table name="sharding_one_1" primaryKey="id" dataNode="dn1" />
		<table name="sharding_one_2" primaryKey="id" dataNode="dn1" />
		<table name="a_two" primaryKey="id" dataNode="dn1,dn2" rule="hash-two" />
        <table name="a_three" primaryKey="id" dataNode="dn1,dn2,dn3" rule="hash-three" />
		<table name="BAMS_FLOW_LOG" primaryKey="id" dataNode="dn1,dn2" rule="hash-two" />
		
		<!--固定Hash分区(string)-->
        <table name="fixed_patch_uniform_string" dataNode="dn1,dn2,dn3,dn4" rule="fixed_uniform_string_rule" />
        <table name="fixed_patch_nonuniform_string" dataNode="dn1,dn2,dn3" rule="fixed_nonuniform_string_rule" />

        <!--枚举方式分区-->
        <table name="enum_patch_integer" dataNode="dn1,dn2,dn3,dn4" rule="enum_integer_rule"/>
		<table name="enum_patch_string" dataNode="dn1,dn2,dn3,dn4" rule="enum_string_rule"/>

		<!--数字范围分区-->
		<table name="range_patch" dataNode="dn1,dn2,dn3,dn4" rule="range_rule" />

        <!--按日期分区-->
		<table name="date_patch" dataNode="dn1,dn2,dn3,dn4" rule="date_rule" />
		<table name="date_patch_default" dataNode="dn1,dn2,dn3,dn4" rule="date_default_rule" />

		<!--全局表-->
		<table name="global_table1" primaryKey="id" type="global" dataNode="dn1,dn2,dn3,dn4" />
		<table name="global_table2" primaryKey="id" type="global" dataNode="dn1,dn2,dn3,dn4" />
		<table name="global_table3" primaryKey="id" type="global" dataNode="dn1,dn2" />
		<table name="test_global" dataNode="dn1,dn2,dn3,dn4" type="global"/>

[...]

PR #1037, opened 2019-03-05
Issue druid1.1.14

yanhuqing666:
Reason:  
   Improve #788 and  Improve lock statement
Type:  
  BUG/Improve  
Influences：  
for #788 
* select * from test_shard where R_NAME=B'1000001';
* lock table test_shard write,test_no_shard write;   [limitation doc](https://github.com/actiontech/dble-docs-cn/blob/develop/3.SQL_Syntax/3.4_Transactional_and_Locking_Statements/3.4.1_Lock&unlock.md#3413-%E9%99%90%E5%88%B6)   
* CREATE TABLE aly_test(id int, data1 varchar(10),data2 varchar(10),data3 varchar(20) GENERATED ALWAYS as (concat(data1,' ',data2))) ;（insert still have problem）
* CREATE TABLE aly_test (id int, data varchar(10), FULLTEXT INDEX idn_id (data));

### DIFF
--- a/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/query/impl/BaseSelectHandler.java
+++ b/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/query/impl/BaseSelectHandler.java
@@ -180,9 +180,9 @@ public class BaseSelectHandler extends BaseDMLHandler {
 
     @Override
     protected void onTerminate() {
-        if (autocommit) {
+        if (autocommit && !session.getSource().isLocked()) {
             this.session.releaseConnection(rrss, LOGGER.isDebugEnabled(), false);
-        } else if (!session.getSource().isLocked()) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/actiontech/dble/backend/mysql/nio/handler/query/impl/BaseSelectHandler.java
```diff
@@ -182,7 +182,7 @@ public void errorResponse(byte[] err, BackendConnection conn) {
     protected void onTerminate() {
         if (autocommit) {
             this.session.releaseConnection(rrss, LOGGER.isDebugEnabled(), false);
-        } else {
+        } else if (!session.getSource().isLocked()) {
```
sunsun314: Would it cause problem when the Complex join with a locked connection?

PR #1039, opened 2019-03-06
fix #1038

### LINKED ISSUE #1038: DDLChildListener#updateMeta should not refresh meta when ddl result is failured.
* **dble version：**  current master branch  Release 2.19.01.0
* **preconditions ：**  
no
* **configs：** 

**schema.xml**  

```



```

**rule.xml**  

```


```

**server.xml**  

```


```

* **steps：**    
    step1.  
```
mysql> create table testpppppp (id int);
Query OK, 0 rows affected (9.91 sec)

mysql> ALTER TABLE `testpppppp` ADD INDEX index_name ( `id` );
Query OK, 0 rows affected (5.95 sec)

mysql> ALTER TABLE `testpppppp` ADD INDEX index_name ( `id` );
ERROR 1061 (42000): Duplicate key name 'index_name
```
* **expect result：**  
    1. When other nodes (not the ddl sql execution node) execute com.actiontech.dble.meta.DDLChildListener#updateMeta , if remote `DDLInfo#status` is `FAILED`, local metaDataLock should be released  but no need to refresh memory metadata.
* **real result：**  
    1. metadata is refreshed when remote DDL sql failed to execute.
image
* **supplements：**  
    1.

/label ~BUG

tankilo:
Reason:  
  BUG
Type:  
  BUG
Influences：  
   fix #1038 
Allow edits from maintainers.

### DIFF
--- a/src/main/java/com/actiontech/dble/meta/DDLChildListener.java
+++ b/src/main/java/com/actiontech/dble/meta/DDLChildListener.java
@@ -89,6 +89,16 @@ public class DDLChildListener implements PathChildrenCacheListener {
             return;
         }
         ClusterDelayProvider.delayBeforeUpdateMeta();
+        // just release local lock
+        if (ddlInfo.getStatus() == DDLStatus.FAILED) {
+            DbleServer.getInstance().getTmManager().removeMetaLock(ddlInfo.getSchema(), table);
+            try {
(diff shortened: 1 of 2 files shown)


src/main/java/com/actiontech/dble/meta/DDLChildListener.java
```diff
@@ -89,6 +89,16 @@ private void updateMeta(ChildData childData) {
             return;
         }
         ClusterDelayProvider.delayBeforeUpdateMeta();
+        // just release local lock
+        if (ddlInfo.getStatus() == DDLStatus.FAILED) {
```
sunsun314: When the node of DDL started execute DDL failed, other node just update the latest table structure. No need to return FAILED flag

Issue #1171, opened 2019-05-10
Discussion about the the logic of MySQLConnection#quit


tankilo:
In method `com.actiontech.dble.backend.mysql.nio.MySQLConnection#quit`, backend connection that passed through `connection-phase` will send a `QUIT` packet to mysql `asynchronouslly`. 
![image](https://user-images.githubusercontent.com/10413284/57496829-8bdd1d00-7307-11e9-8572-22d25bd5e0ae.png)
Because of `NIO` , current thread will not block and continue to close socket. If this is done before 
receiving the reponse of the `QUIT` packet, then when the latter arrvies, `java.nio.channels.ClosedChannelException` is raised.

```
java.nio.channels.ClosedChannelException
  at sun.nio.ch.SocketChannelImpl.ensureReadOpen(SocketChannelImpl.java:257)
  at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:300)
  at com.actiontech.dble.net.NIOSocketWR.asyncRead(NIOSocketWR.java:170)
  at com.actiontech.dble.net.AbstractConnection.asyncRead(AbstractConnection.java:272)
  at com.actiontech.dble.net.NIOReactor$RW.executeKeys(NIOReactor.java:119)
  at com.actiontech.dble.net.NIOReactor$RW.run(NIOReactor.java:90)
  at java.lang.Thread.run(Thread.java:748)
```

What is more serious is that `ResponseHandler#connectionError` will be invoked since `MySQLConnection#quit` doesn't clean `MySQLConnection#respHandler`.  When combined with XA transaction, in method such as `XACommitNodesHandler#errorResponse`，program will 【manually quit backend connection】 by calling `MySQLConnection#quit`. If the above asynchronous problem happend, `XACommitNodesHandler#connectionError` and `XACommitNodesHandler#connectionClose` may be invoked before the response of other datanodes' XA control statement. program will feel there is 【unexpected network failure】happened when executing XA control statement. This will cause concurrent modification of error response message and transaction state, it may lead to problems in latter XA judgement logic.
![image](https://user-images.githubusercontent.com/10413284/57497440-440bc500-730a-11e9-92be-2b06dd7b36ef.png)

![image](https://user-images.githubusercontent.com/10413284/57497955-73bbcc80-730c-11e9-90d5-62e8b415e357.png)

![image](https://user-images.githubusercontent.com/10413284/57497474-69003800-730a-11e9-82bf-2245a6e36122.png)

***
[...]

PR #1175, opened 2019-05-14
Group MySQLConnection close function

sunsun314:
Reason:  
  BUG #1171
Type:  
  Improve  
Influences：  
   Old close function close&quit&terminate group as close&closeWithoutRsp
   Now Socket would be closed right after QUIT packet send out


src/main/java/com/actiontech/dble/backend/mysql/nio/MySQLConnection.java
```diff
@@ -64,7 +64,7 @@
     private long oldTimestamp;
     private final AtomicBoolean logResponse = new AtomicBoolean(false);
     private volatile boolean testing = false;
-
+    private volatile String closeResion = null;
```
yanhuqing666: typo

sunsun314: change to close reason

kaikai2000:
I think, in the NonBlockingSession.terminate() method, the` conn.close` also needs to be rewritten as `conn.closeWithoutRsp()`, otherwise the extra connectionclose() method will be triggered once the serverConnection is closed in a unfinished transaction. Because the serverConnection  has been closed and  then writing the error package will trigger an IoException , and call serverConnection.close() again.
image

Issue #1190, opened 2019-05-20
An alert is required when insert time<start_time on global sequences generated by distributed timestamps 


FlyingMao:
* **dble version：**  
dble-9.9.9.9-96e76d680afe8eed2f618e28cea390b37dde9ffe-20190517071604
* **preconditions ：**  
no
* **configs：** 

**schema.xml**  

```
<table name="mytest_auto_test" dataNode="dn1,dn2,dn3,dn4" rule="hash-four" primaryKey="id" autoIncrement="true"/>
```

**rule.xml**  

```


```

**server.xml**  

```
<property name="sequnceHandlerType">3</property>
```
**sequence_distributed_conf.properties**
```
INSTANCEID=zk
CLUSTERID=01
START_TIME=2019-01-04 09:42:54
```
* **steps：**    
    step1.  Modify system time
```
date -s 2018/01/01
```
  step2.  execute insert in mytest_auto_test table
```
 insert into mytest_auto_test values(1); 
```
* **expect result：**  
    1. Insert failed and give an error
* **real result：**  
    1. insert successfully
* **supplements：**  
    1.

/label ~BUG

FlyingMao:
**verified version：** dble-9.9.9.9-3e92f5e5fde9cb9fb5b328c1d4fd2132e6232a7c-20190524070519

PR #1195, opened 2019-05-21
#1190 - avoid clock moved backwards in case of global sequence generated by zk

Baofengqi:
Reason:  
  BUG #1190. 
Type:  
  BUG
Influences：  
  avoid clock moved backwards in case of global sequence generated by zk


src/main/java/com/actiontech/dble/route/sequence/handler/DistributedSequenceHandler.java
```diff
@@ -202,8 +202,19 @@ public long nextId(String prefixName) throws SQLNonTransientException {
...
         }
-        if (threadLastTime.get() == null) {
-            threadLastTime.set(time);
+        Long lastTimestamp = threadLastTime.get();
+        if (lastTimestamp == null) {
+            if (time >= startTimeMilliseconds) {
+                threadLastTime.set(time);
+                lastTimestamp = time;
+            } else {
+                lastTimestamp = startTimeMilliseconds;
```
sunsun314: lastTimestamp  init when the handler be loaded would be better

Baofengqi: keep it threadlocal

Issue #1231, opened 2019-06-12
SqlJob may finish more than once,and cause other problem


sunsun314:
* **dble version：**  2.19.01.0/rel
* **configs：** 

**schema.xml**  
```
 <schema name="testdb" dataNode="dn1"/>
```
**server.xml**
```
        <property name="checkTableConsistency">0</property>
        <!-- check periodt, he default period is 60000 milliseconds -->
        <property name="checkTableConsistencyPeriod">600000000</property>
        <!-- 1 check the consistency of global table, 0 is not -->
        <property name="useGlobleTableCheck">0</property>
        <property name="glableTableCheckPeriod">86400000</property>

        <!-- heartbeat check period -->
        <property name="dataNodeIdleCheckPeriod">300000000</property>
        <property name="dataNodeHeartbeatPeriod">10000000</property>

```
* **steps：**    
  step1.  stop dble all timed tasks use server.xml
  step2. Add Btrace script delay in com.actiontech.dble.sqlengine.OneRawSQLQueryResultHandler#finished 30s
    Add Btrace script delay in com.actiontech.dble.backend.mysql.nio.MySQLConnection.release 60s
  step3. create table test1(id int,name varchar(20))
  step4. close backconnection in 30s
  step5. wait for the log in Btrace
* **expect result：**  
    SqlJob.finish function sould only be called once
* **real result：**  
    SqlJob.finish can be called more than once
* **supplements：**  
    This issue of SqlJob can cause more problem,such as the tableMeta hang,with exception as below
```
2019-06-11 10:35:08.693  INFO [complexQueryExecutor95] (com.actiontech.dble.backend.mysql.nio.MySQLConnectionHandler.handleDataError(MySQLConnectionHandler.java:233)) - MySQLConnection [id=30, lastTime=1560249308616, user=root, schema=db2, old schema=db2, borrowed=false, fromSlaveDB=false, threadId=3339,character_set_client=utf8,character_set_results=utf8,collation_connection=utf8_general_ci, txIsolation=-1, autocommit=true, attachment=null, respHandler=null, host=172.20.134.2, port=3308, statusSync=null, writeQueue=0, modifiedSQLExecuted=false] handle data error:
java.lang.NullPointerException: null
	at java.util.LinkedList$ListItr.next(LinkedList.java:893) ~[?:1.8.0_121]
[...]

PR #1242, opened 2019-06-24
add list synchronized

sunsun314:
Reason:  
  BUG #1231 
Type:  
  BUG  
Influences：  
   Change the linkedList in Map into synchronized


src/main/java/com/actiontech/dble/meta/table/MultiTablesMetaHandler.java
```diff
@@ -146,7 +146,7 @@ public synchronized void checkTableConsistent(String table, String dataNode, Str
         Map<String, List<String>> tableStruct = tablesStructMap.get(table);
         if (tableStruct == null) {
             tableStruct = new HashMap<>();
-            List<String> dataNodeList = new LinkedList<>();
+            List<String> dataNodeList = Collections.synchronizedList(new LinkedList<>());
```
yanhuqing666: else branch need also use sync to make
  thread safe?

Issue #1261, opened 2019-07-12
MultiSQLJob is not thread safe & countDownShardTable not correct,can cause reload hang


sunsun314:
* **dble version：**  2.19.05.0/rel
* **preconditions ：**  
* **configs：** default config
* **steps：**    
    step1.  create dble tables in detault config 
    step2.  change the created tables' config in schema.xml 
    step3.  add btrace script delay in com.actiontech.dble.sqlengine.MultiSQLJob，line: 155  
    step4.  reload @@config_all   
    step5.  shut down backend mysql

* **expect result：**  
    reload @@config success/or fail  return 
* **real result：**  
    reload @@config hang forever
* **supplements：** 
we can verified on fix version by using btrace script delay in com.actiontech.dble.sqlengine.MultiTablesMetaJob,line 157

sunsun314:
MultiSQLJob  need to be changed to thread safe & countDownShardTable  function should not use just a count to check the reload success

wjl1619:
verified on version: 5.6.29-dble-9.9.9.9-1fcd7adf1b8233d6050245297593240b79eef1e7-20190823072726

PR #1279, opened 2019-07-23
Issue 1261 countDownShardTable not correct,can cause reload hang

sunsun314:
**attention：** THIS FIX IS BASED ON https://github.com/actiontech/dble/pull/1271
Reason:  
  BUG #1261  
Type:  
  BUG  
Influences：  
   Make MultiSQLJob  thread-safe 
    Change & group the handlers in meta reload,make the handlers more improve readability
   countDownShardTable change to set-remove so that the bug of negative-count no longer exist


sunsun314:
will be a new pr later

Issue #1241, opened 2019-06-20
Error,when '-' in the name of schema


FlyingMao:
* **dble version：**  
dble-9.9.9.9-67d04aa94d912fb01583096c6d8cba52f513996e-20190619021040
* **preconditions ：**  
no
* **configs：** 

**schema.xml**  

```
<schema name="schema-1" sqlMaxLimit="100" dataNode="dn5">
		<table name="sharding_4_t1" dataNode="dn1,dn2,dn3,dn4" rule="hash-four" />
</schema>
```

**rule.xml**  

```


```

**server.xml**  

```


```

* **steps：**    
    step1.   execute sql in dble user port
```
mysql> show columns from `schema-1`.`sharding_4_t1`;
ERROR 1146 (42S02): Table 'schema-1.sharding_4_t1' doesn't exist
```
 step2.   execute sql in dble manager port
```
mysql> check full @@metadata where schema='schema-1';
ERROR 1105 (HY000): The sql does not match: check full @@metadata [ where [schema='testSchema' [and table ='testTable']] | [reload_time [>|<]= 'yyyy-MM-dd HH:mm:ss']] | [reload_time is null] | [consistent_in_data_nodes =0|1] | [consistent_in_memory =0|1]
```
* **expect result：**  
    1. execute successful
* **real result：**  
    1. return error，such as steps
* **supplements：**  
    1.

sunsun314:
Well the special characters be supported in schema name and command below are influenced
+ kill @@ddl_lock
+ check full @@metadata
+ reload @@metadata
+ show @@algorithm
+ show @@dataNodes
+ show columns from ..
+ show index from ...
+  show full tables ...
+ show tables status form

FlyingMao:
**verified version：** dble-9.9.9.9-1fcd7adf1b8233d6050245297593240b79eef1e7-20190823072726

sunsun314:
https://github.com/actiontech/dble-docs-cn/blob/develop/2.Function/2.1_manager_cmd/2.1.07_reload.md#2173--reload-metadata

FlyingMao:
**for tester:**
behave_dble/features/func_test/cfg_test/schema.feature#12

PR #1308, opened 2019-08-13
 #1241 support special characters in schema name

sunsun314:
Reason:  
  BUG #1241 
Type:  
  BUG  
Influences：  
the special characters be supported in schema name and command below are influenced
+ kill @@ddl_lock
+ check full @@metadata
+ reload @@metadata
+ show @@algorithm
+ show @@dataNodes
+ show columns from ..
+ show index from ...
+  show full tables ...
+ show tables status form


yanhuqing666:
TIM图片20190814180645

Issue #1300, opened 2019-08-09
Incorrect comment in initSource method of class PhysicalDBPool


ssxlulu:
I think the comment below doesn't show the meaning what it really wants to show. As **'do not creat'** may be changing to **'has created'**.

In PhysicalDBPool#initSource:
```
try {
            if (ds.getTotalConCount() <= 0) {
                ds.initMinConnection(null, true, getConHandler, null);
            } else {
                //In this line, 'do not create' may be changing to 'has created', because the count doesn't less than or equal to zero.
                LOGGER.info("connection with null schema do not create,because testConnection in pool");
                getConHandler.initIncrement();
                hasConnectionInPool = true;
            }
        } catch (Exception e) {
            LOGGER.warn("init connection with schema null error", e);
        }
```
Here is the link to the line:

https://github.com/actiontech/dble/blob/master/src/main/java/com/actiontech/dble/backend/datasource/PhysicalDBPool.java#L384

sunsun314:
Well，if you think the log message is not correct ,you may start a pull request to change it   
We appreciate you every effort to the project

PR #1312, opened 2019-08-14
#1300 Incorrect comment in initSource method of class PhysicalDBPool

ssxlulu:
Reason:  
  Improve #1300 .  
Type:  
  Improve  
Influences：

### DIFF
--- a/src/main/java/com/actiontech/dble/backend/datasource/PhysicalDBPool.java
+++ b/src/main/java/com/actiontech/dble/backend/datasource/PhysicalDBPool.java
@@ -381,7 +381,7 @@ public class PhysicalDBPool {
             if (ds.getTotalConCount() <= 0) {
                 ds.initMinConnection(null, true, getConHandler, null);
             } else {
-                LOGGER.info("connection with null schema has been created,because testConnection in pool");
+                LOGGER.info("connection with null schema has been created,because we tested the connection of the datasource at first");
                 getConHandler.initIncrement();
                 hasConnectionInPool = true;
             }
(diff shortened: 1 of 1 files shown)


src/main/java/com/actiontech/dble/backend/datasource/PhysicalDBPool.java
```diff
@@ -381,7 +381,7 @@ private boolean initSource(int index, PhysicalDatasource ds) {
             if (ds.getTotalConCount() <= 0) {
                 ds.initMinConnection(null, true, getConHandler, null);
             } else {
-                LOGGER.info("connection with null schema do not create,because testConnection in pool");
+                LOGGER.info("connection with null schema has created,because testConnection in pool");
```
Baofengqi: has been ?

ssxlulu: I think that you are right.

src/main/java/com/actiontech/dble/backend/datasource/PhysicalDBPool.java
```diff
@@ -381,7 +381,7 @@ private boolean initSource(int index, PhysicalDatasource ds) {
             if (ds.getTotalConCount() <= 0) {
                 ds.initMinConnection(null, true, getConHandler, null);
             } else {
-                LOGGER.info("connection with null schema do not create,because testConnection in pool");
```
yanhuqing666: can you also fix `because ...`?

ssxlulu: No,I think testConnection is the real cause that created the connection with null schema.

yanhuqing666: > No,I think testConnection is the real cause that created the connection with null schema.

I mean the English grammar... not the reason...

ssxlulu: Okey, could ‘because we tested the connection of the datasource at first’ be appropriate?

yanhuqing666: > Okey, could ‘because we tested the connection of the datasource at first’ be appropriate?

:100:

Issue #759, opened 2018-10-15
expect a way to know whether the config property value is same with the real in working value


irene-coming:
dble version:5.6.29-dble-2.18.10.0-0a33274-20181012083619

Some times, after several times change to config file, I can not tell whether the config value is the one in working. 
so, I expect a way to know config property's in working value

irene-coming:
Ways to get part of config info (provided by [tiger.yan]( https://github.com/yanhuqing666)):
startup time: show @@time.startup
check server.xml property in system node:  show @@sysparam
whitelist: show @@white
schema names in  schema.xml： show @@database
datasource in schema.xml：show @@datasource
datanode in schema.xml： show @@datanode

table info of special schema: show @@dataNodes where schema=? and table=?;
table sharding info: show @@algorithm where schema=? and table=?;

Baofengqi:
add manager command:
show @@user
show @@user.privilege

modify manager command:
show @@datasource

Baofengqi:
update doc : https://github.com/actiontech/dble-docs-cn/blob/develop/2.Function/2.1_manager_cmd/2.1.03_show.md

PR #1334, opened 2019-08-20
avoid npe when schema does't exist in privilege

Baofengqi:
Reason:  
  Improve #759 .  
Type:  
  Improve  
Influences：  
  avoid npe when schema does't exist int privilege

### DIFF
--- a/src/main/java/com/actiontech/dble/manager/response/ShowUserPrivilege.java
+++ b/src/main/java/com/actiontech/dble/manager/response/ShowUserPrivilege.java
@@ -78,20 +78,21 @@ public final class ShowUserPrivilege {
                 continue;
             }
 
-            UserPrivilegesConfig userPrivilegesConfig = user.getPrivilegesConfig();
             String tableName;
+            // privileges
             int[] pri;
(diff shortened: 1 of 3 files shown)


src/main/java/com/actiontech/dble/manager/response/ShowUserPrivilege.java
```diff
@@ -78,20 +78,21 @@ public static void execute(ManagerConnection c) {
...
 
-            UserPrivilegesConfig userPrivilegesConfig = user.getPrivilegesConfig();
             String tableName;
+            // privileges
             int[] pri;
-            if (userPrivilegesConfig == null || !user.getPrivilegesConfig().isCheck()) {
-                for (String schema : user.getSchemas()) {
+            // user privilege config
+            UserPrivilegesConfig userPrivilegesConfig = user.getPrivilegesConfig();
+            boolean isCheck = userPrivilegesConfig == null || !userPrivilegesConfig.isCheck();
```
yanhuqing666: isCheck should be noNeedCheck

Issue #1362, opened 2019-08-29
[imporove]support kill query statement


Baofengqi:
* **dble version：**  
dble-2.19.05.0
* **improve ：**  
support kill query statement
* **document ：**  
https://dev.mysql.com/doc/refman/8.0/en/kill.html

Baofengqi:
doc updated : https://github.com/actiontech/dble-docs-cn/blob/develop/3.SQL_Syntax/3.5_DAL/3.5.3_KILL.md

PR #1363, opened 2019-08-29
add kill query statement

Baofengqi:
Reason:  
  Improve #1362 .  
Type:  
  Improve  
Influences：  
  add kill query statement


Baofengqi:
delay.

Issue #1397, opened 2019-09-19
Wrong Packet Id of MySQL Protocol Packet If Error Happens In Process of Loading Data 


ssxlulu:
I think the packet id is wrong when error hanpens in process of loading data.

> The sequence-id is incremented with each packet and may wrap around. It starts at 0 and is reset to 0 when a new command begins in the Command Phase.

Above is the [MySQL offical document](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_packets.html) about the sequence-id (is same as 'packet id' in this issue).So if the client use 'load data local' synax, and send more than one packet(the packet id will increment one by one), and dble happens error in process of parsing, the err packet dble send will be wrong because the err packet id as i think.

Below is the code in ServerLoadDataInfileHandler#parseFileByLine:
```
try {
                        parseOneLine(row, true);
                    } catch (Exception e) {
                        clear();
                        serverConnection.writeErrMessage(ErrorCode.ER_WRONG_VALUE_COUNT_ON_ROW, "row data can't not calculate a sharding value," + e.getMessage());
                        return false;
                    }
```
As the code show, when error happens the error packet id will be 1. Does it right?

ssxlulu:
The phenomenon when the error happens may be 'Lost connection to MySQL server during query'.

wjl1619:
verified on version: 5.6.29-dble-9.9.9.9-4a1798f8927fa77cef643523b5451807fe469989-20191008111520

PR #1401, opened 2019-09-23
#1397 Wrong Packet Id of MySQL Protocol Packet If Error Happens In Pr…

ssxlulu:
…ocess of Loading Data

Reason:  
  BUG.  
Type:  
  BUG
Influences：

### DIFF
--- a/src/main/java/com/actiontech/dble/server/handler/ServerLoadDataInfileHandler.java
+++ b/src/main/java/com/actiontech/dble/server/handler/ServerLoadDataInfileHandler.java
@@ -208,7 +208,7 @@ public final class ServerLoadDataInfileHandler implements LoadDataInfileHandler 
                 clear();
                 serverConnection.writeErrMessage(ErrorCode.ER_FILE_NOT_FOUND, msg);
             } else {
-                if (parseFileByLine(fileName, loadData.getCharset(), loadData.getLineTerminatedBy())) {
+                if (parseFileByLine((byte) 0, fileName, loadData.getCharset(), loadData.getLineTerminatedBy())) {
                     RouteResultset rrs = buildResultSet(routeResultMap);
                     if (rrs != null) {
                         flushDataToFile();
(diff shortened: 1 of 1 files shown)


src/main/java/com/actiontech/dble/server/handler/ServerLoadDataInfileHandler.java
```diff
@@ -208,7 +208,7 @@ public void start(String strSql) {
                 clear();
                 serverConnection.writeErrMessage(ErrorCode.ER_FILE_NOT_FOUND, msg);
             } else {
-                if (parseFileByLine(fileName, loadData.getCharset(), loadData.getLineTerminatedBy())) {
+                if (parseFileByLine((byte) 0, fileName, loadData.getCharset(), loadData.getLineTerminatedBy())) {
```
sunsun314: input value 0 is improper when MultiQueries would be execute like  
"use db1;load data infile '/opt/test.txt' into table test2"

ssxlulu: As I know,DBLE doesn't truly support the MultiQueries.

As MySQL offical document says about MySQL protocol:
> The sequence-id is incremented with each packet and may wrap around. It starts at 0 and is reset to 0 when a new command begins in the Command Phase.

When a new command begins in the Command Phase, the sequence id will be reset to 0.

ssxlulu: What's more, test must be execute to verfy the theory.

sunsun314: dble has a problem in load data multiQueries. I have created a new issue for the problem https://github.com/actiontech/dble/issues/1403
You can pay attention to or try to fix this problem if you are interested

ssxlulu: Okay,Resolve this problem for simpler scenario first, and i will pay attention to that problem.

Issue #1528, opened 2019-12-04
disable dataHost may cause connection control error


sunsun314:
* **dble version：**  2.19.09.0
* **configs：** 

**schema.xml**  

```
<!DOCTYPE dble:schema SYSTEM "schema.dtd">
<dble:schema xmlns:dble="http://dble.cloud/" version="1.0">
    <schema name="testdb">
        <table name="tb_s" dataNode="dn1" />
    </schema>
    <schema name="testdb2" dataNode="dn1"/>
    <dataNode name="dn1" dataHost="dataHost1" database="db_1"/>
    <dataHost balance="0" maxCon="20" minCon="10" name="dataHost1" switchType="-1" slaveThreshold="100">
        <heartbeat>select 1</heartbeat>
        <writeHost host="hostM1" url="10.186.60.2:3306" password="123456" user="root" disabled="false" id="hostM1" weight="0"/>
    </dataHost>
</dble:schema>

```
**server.xml**  

```
<property name="useOuterHa">true</property>
```

* **steps：**
    +  add debug wait at com.actiontech.dble.backend.datasource.PhysicalDatasource#getConnection
    +  select * from table tb_s
    +  exeucte dataHost @@disable name = 'dataHost1'
    +  repeat for several times
* **expect result：**  
    no special phenomenon
* **real result：**  
    after several times the dataHost connection match the maxCon

yanhuqing666:
merged to master

irene-coming:
there still exists bugs :
1. when dataHost in disabled state, the new query should not create new backend conn, but it does
2.when some to use dataHost in disabled state, the query feedback should tell client about it, not just error like : Backend connect Error, Connection{DataHost[xx.xx.xx.xx:xxxx],Schema[db2]} refused
3.the max conn issue still exists

irene-coming:
verified on dble: 5.6.29-dble-9.9.9.9-4baa5596c340965b8d35e6ee38b81ba9befa196d-20191216104055

PR #1529, opened 2019-12-04
#1528 disable dataHost may cause connection control error

sunsun314:
Reason:  
  BUG #1528 
Type:  
  BUG  
Influences：  
   delete a double check of dataHost disable status, let it check at connection create call back


yanhuqing666:
need to master passed

PR #1677, opened 2020-03-05
Improve：Send heartbeat detection to standby database

gaopengpeng123456:
Reason:  
  Improve ：heartbeat detection of standby database
Type:  
  Improve  
Influences：  
   fix show @@heartbeat;


src/main/java/com/actiontech/dble/backend/datasource/PhysicalDBPool.java
```diff
@@ -343,6 +343,11 @@ public void doHeartbeat() {
         for (PhysicalDatasource source : all) {
             if (source != null) {
                 source.doHeartbeat();
+                for (PhysicalDatasource[] physicalDatasource : source.getDbPool().standbyReadSourcesMap.values()) {
```
sunsun314: standbyReadSourcesMap is a property of this PhysicalDBPool, start the standbyReadSources heartbeat out of the loop to avoid uselessly call

Issue #1688, opened 2020-03-10
[Improve]log4j 2.7 has some safe problem, so upgrade to 2.8.2


Baofengqi:
log4j 2.7 has some safe problem, so upgrade to  2.8.2, including log4j-slf4j-impl, log4j-core and log4j-1.2-api.

PR #1689, opened 2020-03-10
#1688 - upgrade log4j from 2.7 to 2.8.2 due to safe problem

Baofengqi:
Reason:  
  Improve #1688.  
Type:  
  Improve  
Influences：  
  log4j-slf4j-impl, log4j-core and log4j-1.2-api.


Baofengqi:
need more modification

Issue #1706, opened 2020-03-17
NestLoop may cause java.util.ConcurrentModificationException 


uliude:
dble version : 2.19.11/rel
dble with config below:
```
server.xml useJoinStrategy is true
<table name="vs_order_return,vs_store,vs_order" dataNode="dn1,dn2" rule="rule_mod2" />

create table vs_order_return(pk_id int,return_num int,src_order_id int,order_handle_status int)
create table vs_order(pk_id int,sale_act_type int,fk_store_id int)
create table vs_store(pk_id int,fk_org_id int,store_name int)
```
add delay in ：
```
com.actiontech.dble.backend.mysql.nio.handler.builder.HandlerBuilder  Line#51
```
execute sql below:
```
SELECT
		vor.pk_id,
		store.fk_org_id,
		store.store_name storeName,
		vor.return_num orderGoodsNum,
		vo.sale_act_type
FROM
	vs_order_return vor
LEFT JOIN vs_order vo ON vo.pk_id = vor.src_order_id
LEFT JOIN vs_store store ON vo.fk_store_id = store.pk_id
WHERE
	vor.order_handle_status = 3
```
May get error result :
```
Cause: java.sql.SQLException: java.util.ConcurrentModificationException
```
get error stack in dble.log
```
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at com.actiontech.dble.backend.mysql.nio.handler.builder.HandlerBuilder.startHandler(HandlerBuilder.java:50)
	at com.actiontech.dble.backend.mysql.nio.handler.builder.HandlerBuilder.build(HandlerBuilder.java:77)
	at com.actiontech.dble.server.NonBlockingSession.executeMultiResultSet(NonBlockingSession.java:629)
	at com.actiontech.dble.server.NonBlockingSession.executeMultiSelect(NonBlockingSession.java:686)
	at com.actiontech.dble.server.NonBlockingSession.execute(NonBlockingSession.java:490)
	at com.actiontech.dble.server.ServerConnection.routeEndExecuteSQL(ServerConnection.java:341)
	at com.actiontech.dble.server.ServerConnection.execute(ServerConnection.java:279)
	at com.actiontech.dble.server.handler.SelectHandler.handle(SelectHandler.java:108)
[...]

PR #1712, opened 2020-03-18
#1706 NestLoop may cause java.util.ConcurrentModificationException

sunsun314:
Reason:  
  BUG #1706 
Type:  
  BUG  
Influences：  
   NestLoop may cause java.util.ConcurrentModificationException

### DIFF
--- a/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/query/BaseDMLHandler.java
+++ b/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/query/BaseDMLHandler.java
@@ -66,7 +66,7 @@ public abstract class BaseDMLHandler implements DMLResponseHandler {
     }
 
     @Override
-    public List<DMLResponseHandler> getMerges() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/actiontech/dble/backend/mysql/nio/handler/query/BaseDMLHandler.java
```diff
@@ -55,13 +55,18 @@ public final void setNextHandler(DMLResponseHandler next) {
...
+    }
+
     @Override
     public void setLeft(boolean left) {
         this.isLeft = left;
     }
 
     @Override
-    public final List<DMLResponseHandler> getMerges() {
+    public List<DMLResponseHandler> getMerges() {
```
yanhuqing666: why remove final

sunsun314: debug change removed

Issue #1709, opened 2020-03-17
[Improvement] support 16m package or more


gaopengpeng123456:
sql and response package

PR #1726, opened 2020-03-23
16m preparedStatement

gaopengpeng123456:
Reason:  
  Improve #1709
Type:  
  Improve  
Influences：  
   fix 16m  preparedStatement


src/main/java/com/actiontech/dble/net/mysql/CommandPacket.java
```diff
@@ -112,6 +112,53 @@ public void write(MySQLConnection c) {
...
+                c.write(buffer);
+                isFirst = false;
+            }
+            buffer = c.allocate(size + MySQLPacket.PACKET_HEADER_SIZE);
+            BufferUtil.writeUB3(buffer, size);
+            buffer.put(packetId);
+            writeBody(buffer, isFirst, remain);
+            c.write(buffer);
+        } catch (java.nio.BufferOverflowException e1) {
+            //fixed issues #98 #1072
```
sunsun314: does the catch logic meaningful, the code does not use c.allocate() in writeBigPackage() like in write() function

PR #1732, opened 2020-03-24
show @@Heartbeat  response

gaopengpeng123456:
Reason:  
  Improve #1733
Type:  
  Improve  
Influences：  
   fix show @@heartbeat  update rd_code add rs_message


src/main/java/com/actiontech/dble/manager/response/ShowHeartbeat.java
```diff
@@ -135,6 +140,7 @@ public static void response(ManagerConnection c) {
                     row.add(null);
                     row.add(null);
                     row.add(null);
+                    row.add(null);
```
sunsun314: Need one more line here ?

Issue #1676, opened 2020-03-04
Broaden multi-statement support in dble to support 'Set'  multi-statement


sunsun314:
**D**ble support multi-statement in some of the query 'Set' used to not in support list, but we find it normal to use multi-statement like that 'SET ...; SET ...; SELECT....'. So we try to broaden multi-statement support in dble to support 'Set'  multi-statement

PR #1740, opened 2020-03-26
#1676 fix packetId when the sql contains implicit commit or rollback

sunsun314:
Reason:  
  BUG #1676  
Type:  
  BUG  
Influences：  
   fix packetId when the sql contains implicit commit or rollback


sunsun314:
would be fix after xa refactor

Issue #1753, opened 2020-03-31
Add flow control feature to avoid OOM


ssxlulu:
DBLE as a middleware, is responsible for transferring data from backend database or frontend user.

Consider the following scenarios:
**Scenario one**: Transferring data from backend database but the client has stopped received or received as a low speed.
**Scenario two**: Transferring data from client but the backend database recieved as a low speed(We don't consider the stopped receiving scenario because DBLE will cause a exception).

Scenario one will happen when client execute SQL like `SELECT * FROM table`, and Scenario two will happen when client execute `LOAD DATA` SQL.

Both scenarios will cause much memory use in DBLE, even cause OOM. The specific data structure is `writequeue` in `AbstractConnection`.

Flow control feature will control the `writequeue` size to avoid much memory use and OOM exception.

sunsun314:
doc around https://github.com/actiontech/dble-docs-cn/blob/develop/2.Function/2.25_flow_control.md

ssxlulu:
> doc around https://github.com/actiontech/dble-docs-cn/blob/develop/2.Function/2.25_flow_control.md

Good Job!

PR #1758, opened 2020-04-01
Add flow control feature to avoid OOM

ssxlulu:
Reason:  
  Improve #1753.  
Type:  
  Improve  
Influences：  
   Add flow control feature to avoid OOM

### DIFF
--- a/src/main/java/com/actiontech/dble/config/model/SystemConfig.java
+++ b/src/main/java/com/actiontech/dble/config/model/SystemConfig.java
@@ -161,9 +161,9 @@ public final class SystemConfig {
     private ProblemReporter problemReporter;
     private boolean useOuterHa = false;
 
-    private boolean enableFlowControl;
-    private int flowControlStartThreshold;
-    private int flowControlStopThreshold;
+    private boolean enableFlowControl = true;
+    private int flowControlStartThreshold = 4096;
+    private int flowControlStopThreshold = 256;
 
 
     public SystemConfig(ProblemReporter problemReporter) {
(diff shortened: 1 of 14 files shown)


src/main/java/com/actiontech/dble/config/model/SystemConfig.java
```diff
@@ -161,6 +161,11 @@
     private ProblemReporter problemReporter;
     private boolean useOuterHa = false;
 
+    private boolean enableFlowControl = true;
+    private int flowControlStartThreshold = 4096;
```
sunsun314, line 165: the serverConifg param add need to be added in both:
+ server_temp.xml
+ ShowSysParam

Would you like to add them all?

ssxlulu: No problem.

src/main/java/com/actiontech/dble/manager/response/ShowSysParam.java
```diff
@@ -148,6 +148,9 @@ public static void execute(ManagerConnection c) {
         paramValues.add(sysConfig.getMaxCharsPerColumn() + "");
         paramValues.add(sysConfig.getMaxRowSizeToFile() + "");
         paramValues.add(sysConfig.isUseOuterHa() + "");
+        paramValues.add(sysConfig.isEnableFlowControl() + "");
```
sunsun314, line 151: PARAM_NAMES
PARAM_DESCRIPTION
These two object also need to add param into

ssxlulu: Yes, I will add it.

PR #1902, opened 2020-06-28
inner issue 326 shardingNode size = partition size

LUAgam:
Reason:  
  inner issue 326 shardingNode size = partition size
Type:  
  Improve  
Influences：  
  show config error when shardingNode size != partition size


yanhuqing666:
leave it for next version

Issue #1906, opened 2020-07-01
Dble supports the OGG tool for data synchronization


wenyh1:
During the research process of using THE OGG tool to synchronize data to mysql, some sql execution pushed by DBLE is not supported or the result set is empty, 
Such as “Field List table”、“select character_octet_length from information_schema.columns...”
So, we are this compatible with the OGG synchronization process pushed sql, DBLE support OGG data synchronization

PR #1916, opened 2020-07-06
Dble supports the OGG tool for data synchronization 

wd2365151147:
Reason:  
  Improve #1906.  
Type:  
  Improve  
Influences：  
  Dble supports the OGG tool for data synchronization


wd2365151147:
解决冲突

PR #1934, opened 2020-07-09
fix : Cannot generate a path for transactions, view storage, and so on

wd2365151147:
Reason:  
  Improve 
Type:  
  Improve  
Influences：  
   fix : Cannot generate a path for transactions, view storage, and so on


wd2365151147:
Incomplete repair

wd2365151147:
Incomplete repair

PR #2055, opened 2020-08-18
slave instance keeps a heartbeat connection when rwSplitMode == RW_SPLIT_OFF

PanternBao:
Reason:  
  BUG #Inner 465. 
Type:  
  BUG
Influences：  
  slave instance keeps a heartbeat connection when rwSplitMode == RW_SPLIT_OFF


sunsun314 (approved):
OK

PR #2056, opened 2020-08-19
add default values to b variables in python scripts

wenyh1:
Reason:  
  BUG inner 466
Type:  
  BUG
Influences：  
   fix @@enable @@custom_mysql_ha don't take effect


PanternBao (approved):
ok

PR #2057, opened 2020-08-19
set schemaName for table meta when init tableMeta

sunsun314:
Reason:  
  BUG
Type:  
  BUG  
Influences：  
   set schemaName for table meta when init tableMeta to avoid NPException


PanternBao (approved):
lgtm

PR #2060, opened 2020-08-19
improve exception handler in split dump file

PanternBao:
Reason:  
  BUG #inner 469. 
Type:  
  BUG
Influences：  
  improve exception handler in split dump file


LUAgam (approved):
LGTM

Issue #1880, opened 2020-06-11
Framework refactoring & Code Tracing for dble


sunsun314:
Well, dble may add read-write separation or more type function in the future. So that may cause dble change from a "Sharding-only-proxy" to a "Comprehensive-proxy". To fit the challenge dble need a more cleaer service structure.  
Also the tracing function has be mentioned by community users for many times. After the Framework-refactoring it would be easy to add a code tracing into dble, so that community users can use a jaeger UI to observe how dble works

- [x] Structural design
- [x] proto layer
- [x] service layer
- [x] authenticate layer
- [x] tracing function

PR #2065, opened 2020-08-20
fix issue for sql_mode reset

sunsun314:
Reason:  
  BUG #1880   
Type:  
  BUG  
Influences：  
   fix issue for sql_mode reset


LUAgam (approved):
OK

Issue #1994, opened 2020-08-04
when alias contains quota, dble report "cant find column" error


PanternBao:
* **dble version：**  
before 3.20.10.0
* **preconditions ：**  
no
* **configs：** 

**sharding.xml**  

```
<schema name="testdb2" shardingNode="dn1">
        <globalTable name="tb_global2" shardingNode="dn1,dn3"/>
        <shardingTable name="tb_hash_sharding_er2" shardingNode="dn1,dn3" function="func_common_hash"
                       shardingColumn="id"/>
 </schema>       
<function name="func_common_hash" class="Hash">
       <property name="partitionCount">2</property>
       <property name="partitionLength">512</property>
</function>
```
* **steps：**    
    step1.  create table tb_global2(id int, name varchar(64)); create table tb_hash_sharding_er2(id int, name varchar(64));
    step2. select t.name from (select 'test' as "name" from tb_global2) t left join tb_hash_sharding_er2 s on s.id=1;
    step3. select 'test'  from tb_global2 as t inner join tb_hash_sharding_er2 as s on  s.num = t.id;
    step4. select t.code from (select 'code' from tb_global2) t left join tb_hash_sharding_er2 s on s.num=1;
* **expect result：**  
1. step2. successs
2. step3.
 ```
+------+
| test |
+------+
| test |
| test |
| test |
| test |
+------+
4 rows in set (0.01 sec)
```
3. step4.
```
+------+
| code |
+------+
| code |
| code |
| code |
| code |
+------+
4 rows in set (0.01 sec)
```
* **real result：**  
1.step2. ERROR 4004 (42S22): column name not found
2.step3. 
 ```
+--------+
| 'test' |
+--------+
| test   |
| test   |
| test   |
| test   |
+--------+
4 rows in set (0.01 sec)
 ```
3.step4. ERROR 4004 (42S22): column code not found

* **supplements：**  
    1.

quexiuping:
verified on 5.7.21-dble-9.9.9.9-81073d64a76fb91036bbd28a8a932338133ba09c-20201016023638

PR #2068, opened 2020-08-21
solve the problem of quotation marks

wenyh1:
Reason:  
  BUG #1994 
Type:  
  BUG 
Influences：  
   fix: solve the problem of quotation marks


LUAgam (approved):
LGTM

PR #2072, opened 2020-08-25
fix group by related questions

wenyh1:
Reason:  
  BUG inner 473
Type:  
  BUG
Influences：  
   fix group by related questions


yanhuqing666 (approved):
ok

PR #2074, opened 2020-08-25
fix message of the update/delete not allowed && fix readBuffer when t…

sunsun314:
…he buffer is not big enough

Reason:  
  BUG #1880  
Type:  
  BUG
Influences：  
   fix message of the update/delete not allowed && fix readBuffer when the buffer is not big enough


wenyh1 (commented):
LGTM

wenyh1 (approved):

PR #2083, opened 2020-09-01
inner 480

yanhuqing666:
Reason:  
    Improve inner 480
Type:  
  Improve  
Influences：  
   fix desc dble_information.XXX;


wenyh1 (approved):
lgtm

PR #2086, opened 2020-09-02
Dble 475

sunsun314:
Reason:  
  BUG #1880   
Type:  
  BUG  
Influences：  
  fix transformsql to frontend not backend &&add services toString


PanternBao (approved):
ok

PR #2087, opened 2020-09-03
inner 505

wenyh1:
Reason:  
  BUG inner 505
Type:  
  BUG 
Influences：  
   fix  inner-505


wenyh1:
after discussion, the idea of improving is not right

PR #2090, opened 2020-09-04
freshing connection requires add lock

wenyh1:
Reason:  
  BUG inner 351.  
Type:  
  BUG  
Influences：  
   fix inner 351


PanternBao (approved):
lgtm

PR #2094, opened 2020-09-08
fix find bug, lastWriteTime decrement

wenyh1:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  fix find bug, lastWriteTime decrement


wenyh1:
error

PR #2096, opened 2020-09-08
fix bugs find in auto test

sunsun314:
Reason:  
  BUG #1880   
Type:  
  BUG  
Influences：  
   fix bugs find in auto test


yanhuqing666 (approved):
lgtm

PR #2100, opened 2020-09-10
fix for autotest packetId

sunsun314:
Reason:  
  BUG #1880  
Type:  
  BUG  
Influences：  
   fix for autotest packetId, packetId should be set before the handler try to send a errorPacket


yanhuqing666 (approved):
ok

PR #2101, opened 2020-09-10
fix lost of prepareStatement impl

sunsun314:
Reason:  
  BUG #1880   
Type:  
  BUG  
Influences：  
   fix lost of prepareStatement impl


yanhuqing666 (approved):
lgtm

PR #2103, opened 2020-09-11
fix: inner 494/499/500

LUAgam:
Reason:  
  BUG inner 494/499/500
Type:  
  BUG
Influences：  
  fix: inner 494/499/500


PanternBao (approved):
lgtm

PR #2104, opened 2020-09-11
inner-526 join table

yanhuqing666:
Reason:  
  BUG inner-526
Type:  
  BUG 
Influences：  
   join table use handler's self left property


PanternBao (approved):
lgtm

PR #2106, opened 2020-09-14
fix error code in multiNodeQuery

sunsun314:
Reason:  
  BUG #1880   
Type:  
  BUG  
Influences：  
   fix error code in multiNodeQuery


yanhuqing666 (approved):
ok

PR #2110, opened 2020-09-15
fix load balance in sharding mode

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  fix load balance in sharding mode


yanhuqing666 (approved):
LGTM

PR #2116, opened 2020-09-16
fix manager table of session_connections and backend_connections

PanternBao:
Reason:  
  BUG inner 481,491,492,493,527
Type:  
  BUG/Improve  
Influences：  
  fix manager table of session_connections and backend_connections


yanhuqing666 (approved):
lgtm

PR #2118, opened 2020-09-16
change default value of connectionTimeout from 300s to 30s

PanternBao:
Reason:  
   Improve
Type:  
  Improve  
Influences：  
  change value of connectionTimeout from 300s to 30s


yanhuqing666 (approved):
ok

PR #2119, opened 2020-09-16
fix schemaExists is false after reload config

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  fix schemaExists is false after reload config


yanhuqing666 (approved):
ok

PR #2120, opened 2020-09-16
fix:  remove forced conversion

LUAgam:
Reason:  
  BUG inner 533
Type:  
  BUG
Influences：  
  fix:  remove forced conversion


yanhuqing666 (approved):
ok

PR #2121, opened 2020-09-16
fix for xa rollback packetId

sunsun314:
Reason:  
  BUG #1880   
Type:  
  BUG  
Influences：  
   fix for xa rollback packetId


PanternBao (approved):
lgtm

PR #2125, opened 2020-09-17
fix for xa when connection graceful close

sunsun314:
Reason:  
  BUG #1880  
Type:  
  BUG  
Influences：  
   fix for xa when connection graceful close


yanhuqing666 (approved):
lgtm

Issue #2075, opened 2020-08-25
[Improvement] support specialized rwSplit


PanternBao:
support specialized rwSplit

PR #2129, opened 2020-09-17
support prepared statement in rw

PanternBao:
Reason:  
  Improve #2075 .  
Type:  
     Improve  
Influences：  
  support prepared statement in rw


sunsun314 (approved):
LGTM

PR #2131, opened 2020-09-17
fix: autotest

LUAgam:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix: show @@binlog.status only need associated dbgroup


yanhuqing666 (approved):
lgtm

PR #2132, opened 2020-09-17
remove useless log DBLE0REQ-474

sunsun314:
Reason:  
  BUG #1880
Type:  
  BUG  
Influences：  
   remove useless log


PanternBao (approved):
ok

PR #2133, opened 2020-09-17
fix: inner 539/540/541

LUAgam:
Reason:  
  BUG inner 539/540/541
Type:  
  BUG
Influences：  
  fix: inner 539/540/541


PanternBao (approved):
lgtm

Issue #1806, opened 2020-04-16
set multi-user-variables & user variables are dependent result not correct


sunsun314:
* **dble version：**  2.19.11.0/rel
* **preconditions ：**  
* **steps：**    
    step1.  set @a=1;select @a ;set @b=@a+1;select @b
* **expect result：**  
  result b = 2
* **real result：**  
   b = null

yanhuqing666:
delay

wjl1619:
verified on version: drivers/c_mysql_api/multi-queries/mysql_multi_queries.cpp

PR #2136, opened 2020-09-18
fix bug for #1806

sunsun314:
Reason:  
  BUG #1806  
Type:  
  BUG  
Influences：  
   fix set bug when try to set a value for multi-times


yanhuqing666 (approved):
ok

Issue #1987, opened 2020-08-03
[Improve]Add manager schema dble_information and tables to show dble status


yanhuqing666:
You can use the manager port to connect dble and use dble_information.
And use DQL to show the result of memory.
see https://github.com/actiontech/dble-docs-cn/blob/develop/2.Function/2.001_manager_dble_information.md

PR #2137, opened 2020-09-21
add manager table  of processlist

PanternBao:
Reason:  
  Improve #1987.  
Type:  
  Improve  
Influences：  
  add manager table  of processlist


LUAgam (approved):
OK

PR #2139, opened 2020-09-21
add more detail for response service toString

sunsun314:
Reason:  
  BUG #1880  
Type:  
  BUG  
Influences：  
   add more detail for response service toString to fit for the autotest


yanhuqing666 (approved):
ok

PR #2148, opened 2020-09-24
add type for dble slow log

sunsun314:
Reason:  
  Improve slow log
Type:  
  Improve  
Influences：  
   add type for dble slow log


LUAgam (approved):
LGTM

Issue #2147, opened 2020-09-24
select ... information_schema.columns ...; exception when querying with alias


wenyh1:
* **dble version：3.20.07.99**  
* **preconditions ：**  
no
* **configs：** 

**cluster.cnf**  

```


```

**bootstrap.cnf**  

```


```
**db.xml**  

```


```

**user.xml**  

```


```

**sharding.xml**  

```


```

* **steps：**    
    step1.  select column_name as `column_name` from information_schema.columns where table_schema = 'schema_xx' and table_name = 'table_xx';
* **expect result：**  
    1. success
* **real result：**  
    1. ERROR 1003 (HY000): occur Exception, so see dble.log to check reason
See in the log that the issued sql is overappended with 'as column_name', excption message:
```
2020-09-24 05:48:04.074  INFO [complexQueryExecutor3] (com.actiontech.dble.sqlengine.SQLJob.errorResponse(SQLJob.java:129)) - 
error response errNo:1064, You have an error in your SQL syntax; check the manual that corresponds to your MySQL 
server version for the right syntax to use near 'as column_name FROM INFORMATION_SCHEMA.COLUMNS WHERE 
table_schema = 'schema_xx' at line 1 from of sql :SELECT column_name AS `column_name` as column_name FROM 
INFORMATION_SCHEMA.COLUMNS WHERE table_schema = 'schema_xx'AND table_name = 'table_xx'
```
* **supplements：**  
    1.

quexiuping:
verified on 5.7.21-dble-9.9.9.9-81073d64a76fb91036bbd28a8a932338133ba09c-20201016023638

PR #2151, opened 2020-09-25
issue 2147

wenyh1:
Reason:  
  BUG #2147 
Type:  
  BUG
Influences：  
   fix select information_schema.columns' problems with alias, table names


PanternBao (approved):
lgtm

PR #2154, opened 2020-09-27
fix set variables and perfect manager cmd of variables

PanternBao:
Reason:  
  Improve #.  
Type:  
  Improve  
Influences：  
  fix set variables and perfect manager cmd of variables


wenyh1 (commented):
lgmt

wenyh1 (approved):
lgmt

PR #2156, opened 2020-09-28
different rwSplit user can use one rwGroup

PanternBao:
Reason:  
  BUG inner 548
Type:  
  BUG
Influences：  
  different rwSplit user can use one rwGroup


LUAgam (approved):
LGTM

Issue #2153, opened 2020-09-27
In "select ...", when partitionColumn value in "where" is equal to "", causes it to be issued to all nodes


wenyh1:
* **dble version：3.20.07.99**  
* **preconditions ：**  
no
* **configs：** 
sharding table:
CREATE TABLE `test1` (
  `id` int(11) DEFAULT NULL,
  `code` varchar(10) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8

**cluster.cnf**  

```


```

**bootstrap.cnf**  

```


```
**db.xml**  

```


```

**user.xml**  

```


```

**sharding.xml**  

```
<schema name="testdb"  sqlMaxLimit="100">
   <shardingTable name="test1" shardingNode="dn1,dn2" function="func_jumpHash" shardingColumn="id"/>
</schema>

```

* **steps：**    
    step1.  explain select * from test1 where id = '';
* **expect result：**  
    1.
```
+---------------+----------+---------------------------------------------+
| SHARDING_NODE | TYPE     | SQL/REF                                     |
+---------------+----------+---------------------------------------------+
| dn1           | BASE SQL | SELECT * FROM test1 WHERE id = '' LIMIT 100 |
+---------------+----------+---------------------------------------------+
2 rows in set (0.04 sec)
```
* **real result：**  
    1. 
```
+---------------+----------+---------------------------------------------+
| SHARDING_NODE | TYPE     | SQL/REF                                     |
+---------------+----------+---------------------------------------------+
| dn1           | BASE SQL | SELECT * FROM test1 WHERE id = '' LIMIT 100 |
| dn2           | BASE SQL | SELECT * FROM test1 WHERE id = '' LIMIT 100 |
+---------------+----------+---------------------------------------------+
2 rows in set (0.04 sec)
```
* **supplements：**  
    1.

quexiuping:
verified on 5.7.21-dble-9.9.9.9-81073d64a76fb91036bbd28a8a932338133ba09c-20201016023638

PR #2157, opened 2020-09-28
issue 2153

wenyh1:
Reason:  
  BUG #2153 
Type:  
  BUG
Influences：  
   fix when partitionColumn value equal '', forward all nodes


yanhuqing666 (approved):
lgtm

Issue #2162, opened 2020-09-29
Insert select left join is not supported


LUAgam:
* **dble version：3.20.07.0**  
* **preconditions ：**  
no
**sharding.xml**  

```
<shardingTable name="tb_enum_sharding" shardingNode="dn$1-2" sqlMaxLimit="200" function="func_enum"
                       shardingColumn="code"/>
        <shardingTable name="tb_enum_sharding2" shardingNode="dn$1-2" sqlMaxLimit="200" function="func_enum"
                       shardingColumn="code"/>
        <shardingTable name="tb_enum_sharding3" shardingNode="dn$1-2" sqlMaxLimit="200" function="func_enum"
                       shardingColumn="code"/>
```

* **steps：**    
    step1.  
```
drop table if exists tb_enum_sharding;
drop table if exists tb_enum_sharding2;
drop table if exists tb_enum_sharding3;
create table if not exists tb_enum_sharding (
  id int not null,
  code int not null,
  content varchar(250) not null,
  primary key(id)
)engine=innodb charset=utf8;
create table if not exists tb_enum_sharding2 (
  id int not null,
  code int not null,
  content varchar(250) not null,
  primary key(id)
)engine=innodb charset=utf8;
create table if not exists tb_enum_sharding3 (
  id int not null,
  code int not null,
  content varchar(250) not null,
  primary key(id)
)engine=innodb charset=utf8;
insert into tb_enum_sharding values(1,10000,'1'),(2,10010,'2'),(3,10000,'3'),(4,10010,'4');
insert into tb_enum_sharding2 values(1,10000,'1'),(2,10010,'2'),(3,10000,'3'),(4,10010,'4');
```
    step2.  
```
insert into tb_enum_sharding3 select t.* from tb_enum_sharding t left join tb_enum_sharding2 t2 on t.code = t2.code;
```
* **expect result：**  
    1.
```
mysql> insert into tb_enum_sharding3 select t.* from tb_enum_sharding t left join tb_enum_sharding2 t2 on t.code = t2.code;
Query OK, * rows affected (0.07 sec)
```
* **real result：**  
    1.
```
mysql> insert into tb_enum_sharding3 select t.* from tb_enum_sharding t left join tb_enum_sharding2 t2 on t.code = t2.code;
ERROR 1064 (HY000): This `INSERT ... SELECT Syntax` is not supported!
```
* **supplements：**  
    1.

PR #2164, opened 2020-09-30
fix: distinguish outer and inner join

LUAgam:
Reason:  
  BUG #2162
Type:  
  BUG
Influences：  
  fix: distinguish outer and inner join


wenyh1 (approved):
lgtm

Issue #2166, opened 2020-10-09
Failed to execute split caused by semicolon in the string


LUAgam:
* **dble version：2.19.11.99**  
* **preconditions ：**  
no

* **steps：**    
    step1.  Export data through mysqldump, the value of a field in the data contains“;”
    step2.  Call the split command to separate the dump file
* **expect result：**  
    1.success
* **real result：**  
    1.java.sql.SQLSyntaxErrorException: unclosed str. pos 217153, line 1, column 217154, token COMMA
* **supplements：**  
    1.

wjl1619:
verified on version: 5.7.21-dble-9.9.9.9-45370bcf2493a1881e4f9544d4f51697c3d3ff5c-20201020052916

PR #2167, opened 2020-10-09
fix: separator to increase line break mark

LUAgam:
Reason:  
  BUG #2166 
Type:  
  BUG
Influences：  
  fix: separator to increase line break mark


yanhuqing666 (approved):
ok

PR #2169, opened 2020-10-12
fix: distinguish outer and inner join（cherry pick from master #2164）

LUAgam:
Reason:
  BUG #2162
Type:
  BUG
Influences：
  fix: distinguish outer and inner join


PanternBao (approved):
lgtm

PR #2172, opened 2020-10-12
check syn info in backend connection

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  check syn info in backend connection


wenyh1 (approved):
lgtm

PR #2176, opened 2020-10-13
fix: check primary

LUAgam:
Reason:  
  BUG inner 577  
Type:  
  BUG 
Influences：  
  fix: check primary


src/main/java/com/actiontech/dble/services/manager/handler/DeleteHandler.java
```diff
@@ -109,6 +109,7 @@ public void handle(String stmt, ManagerService service) {
                 handleConfigException(e, service, managerTable);
             } else {
                 service.writeErrMessage(ErrorCode.ER_YES, "unknown error:" + e.getMessage());
+                LOGGER.error("unknown error:", e);
```
yanhuqing666, line 112: use WARN ，not error

LUAgam: > use WARN ，not error

ok

wenyh1 (approved):
lgtm

PR #2183, opened 2020-10-15
fix: Session statistics are placed before the task queue

LUAgam:
Reason:  
  BUG inner 578
Type:  
  BUG
Influences：  
  fix: Session statistics are placed before the task queue


yanhuqing666 (approved):
lgtm

PR #2184, opened 2020-10-15
fix: set statement writes to all nodes

LUAgam:
Reason:  
  BUG inner 560
Type:  
  BUG
Influences：  
  fix: set statement writes to all nodes


PanternBao (approved):
lgtm

PR #2185, opened 2020-10-15
fix:inner 575,563

wenyh1:
Reason:  
  BUG inner 575,563
Type:  
  BUG
Influences：  
   fix parameter in any order、return error tips


PanternBao (approved):
lgtm

PR #2188, opened 2020-10-16
support com_change_user in backend

PanternBao:
Reason:  
  Improve #2075.  
Type:  
  Improve  
Influences：  
  support com_change_user in backend

PR #2189, opened 2020-10-16
fix: Print password plugin name

LUAgam:
Reason:  
  Improve inner 579
Type:  
  Improve  
Influences：  
  fix: Print password plugin name


LUAgam:
fixed in #2188

PR #2191, opened 2020-10-19
add default value for processlist

PanternBao:
Reason:  
  BUG #inner 549. 
Type:  
  BUG
Influences：  
  add default value for processlist


wenyh1 (approved):
ok

Issue #2073, opened 2020-08-25
execute sql：“select id,c from test_shard group by c having id<=1 or c = 'bb';” err NPE 


wenyh1:
* **dble version：2.20.04.99**  
* **preconditions ：**  
1.sharding table：create table test_shard(id int,c varchar);
2.'select ...having'  then there's the 'or' condition
* **configs：** 

**schema.xml**  

```
<schema name="schema1" dataNode="dn5">
        <table name="test_shard" dataNode="dn1,dn2,dn3,dn4" rule="hash-four"/>
    </schema>

```

**rule.xml**  

```
no

```

**server.xml**  

```

```

* **steps：**    
    step1.  select id,c from test_shard group by c having id<=1 or c = 'bb';
    step2.  explain select id,c from test_shard group by c having id<=1 or c = 'bb';
* **expect result：**  
    1. similar to mysql: ERROR 1055 (42000): Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'db_1.test_shard.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
    2. 
``` 
+-------------------+-----------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| DATA_NODE         | TYPE            | SQL/REF                                                                                                                                                                        |
+-------------------+-----------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| dn1_0             | BASE SQL        | select `test_shard`.`id`,`test_shard`.`c` from  `test_shard` where (`test_shard`.`id` <= 1) OR (`test_shard`.`c` = 9) GROUP BY `test_shard`.`c` ORDER BY `test_shard`.`c` ASC |
| dn2_0             | BASE SQL        | select `test_shard`.`id`,`test_shard`.`c` from  `test_shard` where (`test_shard`.`id` <= 1) OR (`test_shard`.`c` = 9) GROUP BY `test_shard`.`c` ORDER BY `test_shard`.`c` ASC |
| dn3_0             | BASE SQL        | select `test_shard`.`id`,`test_shard`.`c` from  `test_shard` where (`test_shard`.`id` <= 1) OR (`test_shard`.`c` = 9) GROUP BY `test_shard`.`c` ORDER BY `test_shard`.`c` ASC |
[...]

PR #2192, opened 2020-10-19
fix: supplement where filter

LUAgam:
Reason:  
  BUG inner 590
Type:  
  BUG
Influences：  
  fix: supplement where filter
#2073


wenyh1 (approved):
ok

PR #2194, opened 2020-10-19
fix: supplement where filter(cherry pick from 2.20.04.99 #2192)

LUAgam:
Reason:
  BUG inner 590
Type:
  BUG
Influences：
  fix: supplement where filter
#2073


wenyh1 (approved):
ok

PR #2201, opened 2020-10-23
perf: Improve split log

LUAgam:
Reason:  
  Improve split
Type:  
  Improve  
Influences：  
  perf: Improve split log


LUAgam:
The monitoring method is not good and needs adjustment

PR #2202, opened 2020-10-23
charset for atk-1409, missing for atk-1408

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  charset for atk-1409, missing for atk-1408


yanhuqing666 (approved):
ok

PR #2204, opened 2020-10-26
fix load balance in rwSplit

PanternBao:
Reason:  
  BUG #inner 602.
Type:  
  BUG
Influences：  
  fix load balance in rwSplit


LUAgam (approved):
LGTM

PR #2211, opened 2020-10-28
fix: inner 642

LUAgam:
Reason:  
  BUG inner 642
Type:  
  BUG
Influences：  
  fix: inner 642
        auth switch request should use client configuration


PanternBao (approved):
ok

PR #2213, opened 2020-10-29
when set value is binary, turn the value to hex string stored

PanternBao:
Reason:  
  BUG #inner 609,611 
Type:  
  BUG
Influences：  
  when set value is binary, turn the value to hex string stored


LUAgam (approved):
LGTM

PR #2214, opened 2020-10-29
fix chinese in column and row

PanternBao:
Reason:  
  BUG
Type:  
  BUG 
Influences：  
    fix chinese in column and row


yanhuqing666 (approved):
lgtm

PR #2215, opened 2020-10-29
fix: inner 645 and add log

LUAgam:
Reason:  
  BUG inner 645 
Type:  
  BUG
Influences：  
  fix: inner 645 and add log
auth switch request should use mysql_native_password


PanternBao (approved):
ok

PR #2220, opened 2020-10-30
support expression in set statement

PanternBao:
Reason:  
  BUG #inner 635.
Type:  
  BUG
Influences：  
  support expression in set statement


LUAgam (approved):
LGTM

PR #2223, opened 2020-11-02
don't load sharding.xml when only rwSplit user is configured

PanternBao:
Reason:  
  BUG #inner 639.
Type:  
  BUG 
Influences：  
  don't load sharding.xml when only rwSplit user is configured


yanhuqing666 (approved):
ok

PR #2224, opened 2020-11-02
avoid the following case: 1. login 2. set autocommit=0 3. select xxx

PanternBao:
Reason:  
  BUG #inner-591
Type:  
  BUG
Influences：  
  avoid the following case: 1. login 2. set autocommit=0 3. select xxx


PanternBao:
The branch was deleted by mistake.

LUAgam (approved):
LGTM

PR #2226, opened 2020-11-03
fix: inner 648

LUAgam:
Reason:  
  BUG inner 648
Type:  
  BUG
Influences：  
  fix: inner 648
when auth request is \004 need send switch auth response


PanternBao (approved):
ok

PR #2228, opened 2020-11-03
data type is binary, need check flag to judge whether the data is number or string

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
   fix ci - data type is binary, need check flag to judge whether the data is number or string


LUAgam (approved):
LGTM

PR #2231, opened 2020-11-04
fix: inner 650

LUAgam:
Reason:  
  BUG inner 650
Type:  
  BUG
Influences：  
  fix: inner 650
Add packet number when unpacking


wenyh1 (approved):
ok

PR #2233, opened 2020-11-04
fix bug that idleTimeout does not take effect

PanternBao:
Reason:  
  BUG #inner654.
Type:  
  BUG
Influences：  
  fix bug that idleTimeout does not take effect


yanhuqing666 (approved):
ok

Issue #2152, opened 2020-09-27
column after 'where' does not specify the table name，causes it to be issued to all nodes


wenyh1:
* **dble version：3.20.07.99**  
* **preconditions ：**  
tables:
1.drop table if exists tb_parent;
create table if not exists tb_parent (
  id int not null,
  id2 int not null,
  content varchar(250) not null,
  primary key(id)
)engine=innodb charset=utf8;
insert into tb_parent values(1,1,'1'),(2,2,'2'),(513,513,'513');
2.drop table if exists tb_child2;
create table if not exists tb_child2 (
  child2_id int not null,
  content varchar(250) not null
)engine=innodb charset=utf8;
insert into tb_child2 values(1,'1');
insert into tb_child2 values(2,'2');
insert into tb_child2 values(513,'513');

* **configs：** 

**cluster.cnf**  

```


```

**bootstrap.cnf**  

```


```
**db.xml**  

```


```

**user.xml**  

```


```


**sharding.xml**  

```
<shardingTable name="tb_parent" shardingNode="dn1,dn2" function="func_jumpHash" shardingColumn="id">
    <childTable name="tb_child2" joinColumn="child2_id" parentColumn="id"/>
</shardingTable>

<function name="func_jumpHash" class="jumpStringHash">
    <property name="partitionCount">2</property>
    <property name="hashSlice">0:2</property>
</function>

```

* **steps：**    
    step1.  execute the following sql(notes here 'id' after 'where' does not specify the table name):
``` sql
explain select * from tb_parent inner join tb_child2  on tb_child2.child2_id = tb_parent.id where id = 0;
```

* **expect result：**  
    1.
```
+---------------+----------+------------------------------------------------------------------------------------------------------------+
| SHARDING_NODE | TYPE     | SQL/REF                                                                                                    |
+---------------+----------+------------------------------------------------------------------------------------------------------------+
| dn1           | BASE SQL | select * from tb_parent inner join tb_child2  on tb_child2.child2_id = tb_parent.id where tb_parent.id = 0 |
+---------------+----------+------------------------------------------------------------------------------------------------------------+
1 row in set (0.01 sec)
```
* **real result：**  
    1.
```
+-----------------+---------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| SHARDING_NODE   | TYPE          | SQL/REF                                                                                                                                                                                                                    |
[...]

PR #2235, opened 2020-11-04
issue 2152

wenyh1:
Reason:  
  BUG #2152  
Type:  
  BUG
Influences：  
   fix :not specify the table name，causes to all nodes


yanhuqing666 (approved):
LGTM

PR #2239, opened 2020-11-05
fix: inner 637

LUAgam:
Reason:  
  BUG inner 637
Type:  
  BUG
Influences：  
  fix: inner 637

### DIFF
--- a/src/main/java/com/actiontech/dble/services/manager/response/ShowBackend.java
+++ b/src/main/java/com/actiontech/dble/services/manager/response/ShowBackend.java
@@ -213,14 +213,14 @@ public final class ShowBackend {
         row.add(c.getCharsetName().getClient().getBytes());
         row.add(c.getCharsetName().getCollation().getBytes());
         row.add(c.getCharsetName().getResults().getBytes());
-        row.add(c.isClosed() ? new byte[]{0} : (c.getBackendService().getTxIsolation() + "").getBytes());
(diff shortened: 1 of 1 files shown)


src/main/java/com/actiontech/dble/services/manager/response/ShowBackend.java
```diff
@@ -213,13 +213,15 @@ private static RowDataPacket getRow(BackendConnection c, String charset) {
         row.add(c.getCharsetName().getClient().getBytes());
         row.add(c.getCharsetName().getCollation().getBytes());
         row.add(c.getCharsetName().getResults().getBytes());
-        row.add((c.getBackendService().getTxIsolation() + "").getBytes());
-        row.add((c.getBackendService().isAutocommit() + "").getBytes());
-        row.add(StringUtil.encode(c.getBackendService().getStringOfSysVariables(), charset));
-        row.add(StringUtil.encode(c.getBackendService().getStringOfUsrVariables(), charset));
-        row.add(StringUtil.encode(c.getBackendService().getXaStatus().toString(), charset));
+        row.add(c.isClosed() ? new byte[]{0} : (c.getBackendService().getTxIsolation() + "").getBytes());
```
yanhuqing666, line 216: what's the difference between `new byte[]{0} ` and `null`

LUAgam: > what's the difference between `new byte[]{0} ` and `null`

When the connection is closed, the relationship between connection and service is released

wenyh1 (approved):
lgtm

PR #2242, opened 2020-11-05
when query SQL exist unknown schema, npe error

wenyh1:
Reason:  
  BUG inner 627
Type:  
  BUG
Influences：  
   fix : when query SQL exist unknown schema, npe error


PanternBao (approved):
lgtm

PR #2250, opened 2020-11-05
fix: inner 652(cherry pick from master #2237)

LUAgam:
Reason:
  BUG inner 652
Type:
  BUG
Influences：
  fix: inner 652
#2237


PanternBao (approved):
ok

PR #2252, opened 2020-11-05
fix: ci bug

wenyh1:
Reason:  
  BUG inner 662
Type:  
  BUG
Influences：  
   1.fix  inner 662
        2.execute "show @@dbinstance where shardingNode = 'dn1'"  cause npe


yanhuqing666 (approved):
lgtm

PR #2257, opened 2020-11-06
fix MySQLFrontAuthService cast to FrontEndService fail inner DMP-5863

ylinzhu:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

Issue #1851, opened 2020-05-12
sharing_table in view will cause file import from `split` to fail


FlyingMao:
* **dble version：**  
dble-2.20.04.99-a52c31b01ece0f248bdb31890bc2bdf268173ae2-20200509072401
* **preconditions ：**  
no
* **configs：** 

**schema.xml**  

```

<schema name="schema1" sqlMaxLimit="100" dataNode="dn5">
        
		<table name="sharding_4_t1" dataNode="dn1,dn2,dn3,dn4" rule="hash-four" incrementColumn="id"  />

	</schema>
	<schema name="schema2" sqlMaxLimit="100">
    <dataNode name="dn1" dataHost="10.186.60.41" database="db1"/>
    <dataNode name="dn2" dataHost="10.186.60.42" database="db1"/>
    <dataNode name="dn3" dataHost="10.186.60.41" database="db2"/>
    <dataNode name="dn4" dataHost="10.186.60.42" database="db2"/>
     <dataNode name="dn5" dataHost="10.186.60.41" database="db3"/>

```

**rule.xml**  

```


```

**server.xml**  

```


```

* **steps：**    
    step1.  execute sql in mysql
```
use schema1;
create table sharding_4_t1(id int primary key AUTO_INCREMENT,age int,pid int);
create view view_test as select * from sharding_4_t1;
```
  step2.  dump
```
mysqldump -utest -ptest -h127.0.0.1 -P3306 --databases schema1 --set-gtid-purged=OFF > /opt/mysql_dump_large.sql
```
   step3. execute sql in dble manager port
```
split /opt/mysql_dump_large.sql /opt/;
```
  step4. import the file from `split` to dn5
```
mysql3306 -D db3<mysql_dump_large.sql-dn5-1589263782239.dump
```
* **expect result：**  
    1. successed
* **real result：**  
    1.
```
[root@localhost opt]# mysql3306 -D db3<mysql_dump_large.sql-dn5-1589263782239.dump 
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1146 (42S02) at line 105: Table 'db3.sharding_4_t1' doesn't exist
```
* **supplements：**  
    1.
```
[root@localhost opt]# cat mysql_dump_large.sql |grep view
-- Temporary view structure for view `view_test`
DROP TABLE IF EXISTS `view_test`;
/*!50001 DROP VIEW IF EXISTS `view_test`*/;
[...]

PR #2258, opened 2020-11-06
fix bug that the split file contains view definition

PanternBao:
Reason:  
  BUG #1851 .
Type:  
  BUG 
Influences：  
    fix bug that the split file contains view definition


LUAgam (approved):
lgtm

PR #2260, opened 2020-11-09
fix: inner 648

LUAgam:
Reason:
  BUG inner 648
Type:
  BUG
Influences：
  fix: inner 648
Heartbeat detection distinguishes internal retries and sub-thread retries


LUAgam:
Internal retry and background retry may be concurrent, but the probability is low and does not affect the final result. Observe whether there are other problems in subsequent tests

LUAgam:
not reproduced at present, wait for subsequent reappear and then repair

PR #2261, opened 2020-11-09
fix: inner 670

LUAgam:
Reason:  
  BUG inner 670
Type:  
  BUG
Influences：  
  fix: inner 670
ArrayList thread unsafe


yanhuqing666 (approved):
lgtm

PR #2263, opened 2020-11-10
fix: inner 670(cherry pick from master #2262) 

LUAgam:
Reason:  
  BUG inner 670
Type:  
  BUG
Influences：  
  fix: inner 670
Hang caused by the specified non-existent schema in the connection
#2262


yanhuqing666 (approved):
lgtm

PR #2268, opened 2020-11-11
fix bug that dble can't start in performance mode

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  fix hang in performance mode


yanhuqing666 (approved):
ok

PR #2269, opened 2020-11-11
fix inner DBLE0REQ-607

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   support begin work and commit work


LUAgam (approved):
ok

PR #2271, opened 2020-11-11
fix: inner 674 

LUAgam:
Reason:  
  BUG inner 674 
Type:  
  BUG
Influences：  
  fix: inner 674 
distinguish shardingUseless and rwSplitUseless


PanternBao (approved):
lgtm

PR #2272, opened 2020-11-11
fix bug about kill query

PanternBao:
Reason:  
  BUG #inner 649, inner651.
Type:  
  BUG
Influences：  
  fix bug about kill query


LUAgam (approved):
lgtm

PR #2274, opened 2020-11-11
client lost connection when all sharding node can't connect

PanternBao:
Reason:  
  BUG #inner680
Type:  
  BUG
Influences：  
  client lost connection when all sharding node can't connect


yanhuqing666 (approved):
ok

PR #2275, opened 2020-11-12
fix NullPointException in use performance mode

PanternBao:
Reason:  
  BUG.  
Type:  
  BUG
Influences：  
  fix NullPointException in use performance mode


yanhuqing666 (approved):
lgtm

PR #2277, opened 2020-11-12
fix: inner 678

LUAgam:
Reason:  
  BUG inner 678
Type:  
  BUG
Influences：  
  fix: inner 678
* npe
* solve packet disorder
* the value of the disabled field of the DbleDbInstance table is incorrect


yanhuqing666 (approved):
lgtm

PR #2279, opened 2020-11-13
fix bug in performance mode

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  fix bug in performance mode


PanternBao:
resolve conficts

PR #2291, opened 2020-11-18
fix: service is set to null earlier than connection close(cherry pick from master #2290)

LUAgam:
Reason:
  BUG ci
Type:
  BUG
Influences：
  fix: service is set to null and connection close is not atomic
service is set to null earlier than connection close
#2290


wenyh1 (approved):
ok

PR #2292, opened 2020-11-18
fix: Miss bug

wenyh1:
Reason:  
  BUG
Type:  
  BUG
Influences：  
   fix when query dual table and schema is null, NPE Exception


PanternBao (approved):
ok

PR #2295, opened 2020-11-18
correct size to maxCon in show @@dbInstance

PanternBao:
Reason:
  BUG #inner 703.
Type:
  BUG
Influences：
  correct size to maxCon in show @@dbInstance


wenyh1 (approved):
ok

PR #2298, opened 2020-11-19
modify tips when there are no rwsplit user or sharding user in dryrun

PanternBao:
Reason:  
  Improve #inner708.  
Type:  
  Improve  
Influences：  
  modify tips when there are no rwsplit user or sharding user in dryrun


yanhuqing666 (approved):
ok

PR #2301, opened 2020-11-20
 inner DBLE0REQ-657 (cherry pick from 3.20.10 #2270)

ylinzhu:
Reason:  
  BUG  
Type:  
  BUG
Influences：  
   delete semicolons if last position of sql have it


wenyh1 (approved):
ok

PR #2302, opened 2020-11-20
'select...for update' and 'select...lock in share mode' sql, to master node

wenyh1:
Reason:  
  BUG #inner-704 
Type:  
  BUG  
Influences：  
  improve 'select...for update' and 'select...lock in share mode' sql, to master node


yanhuqing666:
I think the broadcast of `select... for update`  will hang

wenyh1:
It will be revised later

PR #2306, opened 2020-11-20
add primary key for manager table

PanternBao:
Reason:  
   Improve.  
Type:  
      Improve  
Influences：  
   add primary key for manager table


wenyh1 (approved):
ok

PR #2307, opened 2020-11-20
fix: when character_set_results is null,the default value will be used binary

LUAgam:
Reason:
  BUG inner 710
Type:
  BUG
Influences：
  fix: when character_set_results is null,the default value will be used binary


yanhuqing666 (approved):
lgtm

PR #2308, opened 2020-11-21
'select...for update' and 'select...lock in share mode' sql, to master node

wenyh1:
Reason:  
  BUG inner-704
Type:  
  BUG
Influences：  
   Improve   'select...for update' and 'select...lock in share mode' sql, to master node


PanternBao (approved):
lgtm

PR #2312, opened 2020-11-23
Fixed some StringIndexOutOfBoundsException when sql parse.

dcy10000:
Reason:  
  Some special string in the manage SQL parser may cause  StringIndexOutOfBoundsException.
Type:  
  BUG
Influences：  
  Fixed some StringIndexOutOfBoundsException.


yanhuqing666 (approved):
ok

PR #2315, opened 2020-11-24
fix inner DBLE0REQ-607 support begin work to RWsplit (cherry pick from master#2269)

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
fix inner DBLE0REQ-607 support begin work to RWsplit (cherry pick from master#2269)


LUAgam (approved):
lgtm

PR #2316, opened 2020-11-24
fix inner DBLE0REQ-698 support help keyword load balance mode

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG 
Influences：  
  fix inner DBLE0REQ-698 support help keyword load balance mode


PanternBao (approved):
ok

PR #2317, opened 2020-11-24
fix: release and faked timing adjustment

LUAgam:
Reason:  
  BUG inner 652
Type:  
  BUG
Influences：  
  fix: release and faked timing adjustment
inner 652


PanternBao (approved):
lgtm

PR #2318, opened 2020-11-24
avoid the case that the packet sequence is field,eof,err when executing select

PanternBao:
Reason:  
  BUG #inner 709. 
Type:  
  BUG
Influences：  
  avoid the case that the packet sequence is field,eof,err when executing select


yanhuqing666 (approved):
lgtm

PR #2320, opened 2020-11-24
fix: slave dble reload failure

wenyh1:
Reason:  
  BUG inner DMP-6128
Type:  
  BUG
Influences：  
   fix: slave dble reload failure


PanternBao (approved):
lgtm

PR #2321, opened 2020-11-25
fix: whiteIp wildcard modification

LUAgam:
Reason:  
  BUG inner 712
Type:  
  BUG
Influences：  
  fix: whiteIp wildcard modification
inner 712


PanternBao (approved):
lgtm

PR #2325, opened 2020-11-25
inner DMP-6115 fixed: show which file cause error in dry-run.

dcy10000:
Reason:  
  BUG inner-DMP-6115
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #2326, opened 2020-11-25
fix: whiteIp distinguish fe80.%,wildcard 

LUAgam:
Reason:  
  BUG inner 714
Type:  
  BUG
Influences：  
  fix: whiteIp distinguish fe80.%,wildcard 
inner 714


wenyh1 (approved):
lgtm

PR #2327, opened 2020-11-25
fix inner DBLE0REQ-698 support help keyword load balance mode(cherry-pick from 3.20.10.99 #2316)

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG 
Influences：  
   fix inner DBLE0REQ-698 support help keyword load balance mode


LUAgam (approved):
lgtm

PR #2330, opened 2020-11-26
fix: set the key of traceResult to connection id+node name+sql

LUAgam:
Reason:  
  BUG inner 676
Type:  
  BUG
Influences：  
  fix: set the key of traceResult to connection id+node name+sql


PanternBao (approved):
lgtm

PR #2332, opened 2020-11-27
Inner 721 no sharding/split user

yanhuqing666:
Reason:  
    Improve   
Type:  
   Improve  
Influences：  
   no sharding/split user dryrun level


PanternBao (approved):
lgtm

PR #2333, opened 2020-11-27
fix: the priority table of sqlMaxLimit is greater than the schema

LUAgam:
Reason:  
  BUG inner 720
Type:  
  BUG
Influences：  
  fix: The priority table of sqlMaxLimit is greater than the schema


wenyh1 (approved):
ok

PR #2335, opened 2020-11-27
fix: release and faked timing adjustment(cherry pick from master #2317)

LUAgam:
Reason:
  BUG inner 652
Type:
  BUG
Influences：
  fix: release and faked timing adjustment(cherry pick from master #2317)


PanternBao (approved):
lgtm

PR #2337, opened 2020-11-27
change primary key from front_id,sharding_node to front_id,mysql_id in processlist

PanternBao:
Reason:  
  BUG
Type:  
  BUG  
Influences：  
  change primary key from front_id,sharding_node to front_id,mysql_id in processlist


LUAgam (approved):
ok

PR #2341, opened 2020-11-27
inner-726 kill return OK.But doesn't get any effect.

dcy10000:
Reason:  
  BUG  726
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #2343, opened 2020-11-27
inner-722: detect shardingNode repeat

dcy10000:
Reason:  
  BUG 722
Type:  
  BUG/Improve  
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #2344, opened 2020-11-30
add log observation

wenyh1:
Reason:  
  BUG inner DMP-6236
Type:  
  BUG
Influences：  
   no fix, add log observation


PanternBao (approved):
lgtm

PR #2345, opened 2020-11-30
distributed locks not being released

wenyh1:
Reason:  
  BUG DMP-6183
Type:  
  BUG
Influences：  
   fix: distributed locks not being released


yanhuqing666 (approved):
ok

PR #2352, opened 2020-12-01
inner-358: expect to provide "show dble_processlist" in server port

dcy10000:
Reason:  
     inner-358
Type:  
  Improve  
Influences：  
   fix xx


LUAgam (approved):
lgtm

PR #2354, opened 2020-12-01
fix: the priority table of sqlMaxLimit is greater than the schema(cherry pick from master #2333)

LUAgam:
Reason:
  BUG inner 720
Type:
  BUG
Influences：
  fix: the priority table of sqlMaxLimit is greater than the schema(cherry pick from master #2333)


wenyh1 (approved):
ok

PR #2357, opened 2020-12-01
atk-1489:improve er left join global

PanternBao:
Reason:  
   Improve #atk-1489.  
Type:  
  Improve  
Influences：  
  improve er left join global


LUAgam (approved):
lgtm

PR #2361, opened 2020-12-02
resume can be executed on other dble nodes

wenyh1:
Reason:  
  BUG
Type:  
  BUG
Influences：  
   resume can be executed on other dble nodes


yanhuqing666 (approved):
lgtm

PR #2365, opened 2020-12-07
fix: 't_' environment variable was referenced... in wrapper.log

wenyh1:
Reason:  
  Improve inner-772
Type:  
  Improve  
Influences：  
   fix: 't_' environment variable was referenced... in wrapper.log


PanternBao (approved):
lgtm

PR #2367, opened 2020-12-07
inner-773：remove "/" and wrap in error report when reload config.

dcy10000:
Reason:  
  BUG inner-773
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #2370, opened 2020-12-08
inner-722-supplement: modify description

dcy10000:
Reason:  
   Improve inner-722
Type:  
  Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #2372, opened 2020-12-09
'-javaagent' configuration item is supported in the bootstrap.cnf

wenyh1:
Reason:  
  BUG inner 789
Type:  
  BUG
Influences：  
  '-javaagent' configuration item is supported in the bootstrap.cnf


PanternBao (approved):
lgtm

PR #2374, opened 2020-12-14
fix: when the password is updated, the problem of encoding and decoding

LUAgam:
Reason:  
  BUG inner 797
Type:  
  BUG
Influences：  
  fix: when the password is updated, the problem of encoding and decoding


PanternBao (approved):
lgtm

PR #2375, opened 2020-12-14
 fix: not in line with expectation of routing to single node sql

wenyh1:
Reason:  
  BUG inner 732
Type:  
  BUG
Influences：  
   fix: not in line with expectation of routing to single node sql


LUAgam (approved):
lgtm

PR #2378, opened 2020-12-15
fix: problems with set names and set charset

LUAgam:
Reason:  
  BUG inner 603/605/606
Type:  
  BUG
Influences：  
  fix: problems with set names and set charset
1.Ensure that the input sql character set of dble and mysql is consistent
2.This statement maps all strings sent between the server and the current client with the given mapping. SET CHARACTER SET sets three session system variables: character_set_client and character_set_results are set to the given character set, and character_set_connection to the value of character_set_database.
ref:https://dev.mysql.com/doc/refman/5.7/en/set-character-set.html


wenyh1 (approved):
lgtm

PR #2379, opened 2020-12-17
inner-557: fixed: temporary table query should send to master

dcy10000:
Reason:  
  In RwSplit scenarios, the `create temporary table` query doesn't send to the master. But the 'select from temporary table' query send to slave. So cause `table not found ` error.
Type:  
  BUG  inner 557
Influences：  

- If I create a temporary table, all queries would send to master unless drop all temporary tables afterward.


PanternBao (approved):
lgtm

dcy10000:
associated with pr https://github.com/actiontech/dble/pull/2395

PR #2380, opened 2020-12-17
Inner 711&631: fixed:multi-query cause timeout and  'CALL' query causes hang.

dcy10000:
Reason:  
  In RwSplit scenarios, when using JDBC with URL param `allowMultiQueries`, send multi-query cause timeout.Error reports:`(2013, 'Lost connection to MySQL server during query')`
Type:  
  BUG inner-711
Influences：  
   fixed.

Reason:  
  In RwSplit scenarios, the `CALL` query causes hang.
Type:  
  BUG inner-631
Influences：  
   fixed.

### DIFF
--- a/src/main/java/com/actiontech/dble/server/parser/RwSplitServerParse.java
+++ b/src/main/java/com/actiontech/dble/server/parser/RwSplitServerParse.java
@@ -130,8 +130,8 @@ public final class RwSplitServerParse extends ServerParse {
 
 
     public static boolean isMultiStatement(String sql) {
-        final int index = sql.indexOf(';');
-        return index != -1 && index != sql.length() - 1;
+        int index = ParseUtil.findNextBreak(sql);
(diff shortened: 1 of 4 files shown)


src/main/java/com/actiontech/dble/server/parser/RwSplitServerParse.java
```diff
@@ -128,6 +128,12 @@ public static int parse(String stmt) {
         return rt;
     }
 
+
+    public static boolean isMultiStatement(String sql) {
+        final int index = sql.indexOf(';');
```
LUAgam, line 133: whether to consider value of insert has ';'

LUAgam (approved):

src/main/java/com/actiontech/dble/server/parser/RwSplitServerParse.java
dcy10000: No, I will fix it right away.

LUAgam (approved):
lgtm

PR #2383, opened 2020-12-28
When the character set for the client does not exist in dble, the default character set for the corresponding version of FakeMySQLVersion in dble is used

wenyh1:
Reason:  
  BUG inner 355
Type:  
  BUG
Influences：  
   fix


LUAgam (approved):
lgtm

PR #2385, opened 2020-12-30
fix : the database does not exist in mysql but can be logged in rwsplituser inner 668 and 667 

ylinzhu:
fix : the database does not exist in mysql but can be logged in rwsplituser inner 668 and 667 

Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix : the database does not exist in mysql but can be logged in rwsplituser inner 668 and 667


src/main/java/com/actiontech/dble/services/mysqlauthenticate/util/AuthUtil.java
```diff
@@ -175,6 +187,17 @@ private static int checkSchema(String schema, ShardingUserConfig userConfig) {
         }
     }
 
+    private static int checkDatabase(String database) {
```
wenyh1, line 190: whether to consider case sensitivity？

ylinzhu:
add sensitivity check

wenyh1 (approved):
ok

PR #2390, opened 2021-01-04
inner-826 : fixed drop table doesn't take any effect

dcy10000:
Reason:  
  BUG  inner-826
Type:  
  BUG
Influences：  
   server parse about drop.


wenyh1 (approved):
lgtm

PR #2391, opened 2021-01-05
inner-747: fixed: dble's HEX() should be the same as  mysql's

dcy10000:

Reason:  
  BUG inner-747
Type:  
  BUG
Influences：  
   ItemFunc


LUAgam (approved):
lgtm

PR #2393, opened 2021-01-05
inner-756: adjust quote output

dcy10000:

Reason:  
  BUG inner-756
Type:  
  BUG
Influences：  
   quote func


LUAgam (approved):
ok

PR #2395, opened 2021-01-05
inner-557-supplement: ServerParse using repeat id

dcy10000:

Reason:  
       supplement to https://github.com/actiontech/dble/pull/2379 .
  
       Used repeat id. Although cause no error yet. But it is a bad practice.
Type:  
  Improve  
Influences：  
   parse sql.


wenyh1 (approved):
ok

PR #2396, opened 2021-01-05
fix: the mysql instance version cannot be lower than the dble version

LUAgam:
Reason:  
  BUG inner 647
Type:  
  BUG
Influences：  
the mysql instance version cannot be lower than the dble version
the charsetutil reference does not work due to the timing of fakeMySQLVersion value assignment


wenyh1 (approved):
lgtm

PR #2397, opened 2021-01-06
inner-730: fixed: 'use schema'  cause slowQuery  thread terminated.

dcy10000:

Reason:  
  BUG inner-730
Type:  
  BUG
Influences：  
   1.show tables
   2.write slowQuery log


PanternBao (approved):
lgtm

PR #2399, opened 2021-01-06
feat: JSON mode configuration takes effect

LUAgam:
Reason:  
  Improve #2386   inner-453/658/507
Type:  
  Improve 
Influences：  
Cluster subscriber update configuration
insert, delete and update of dble_information (single machine, cluster)


LUAgam:
Needs to be changed

PR #2404, opened 2021-01-07
check heartbeat status when dble select dbInstance

PanternBao:
Reason:  
  BUG #inner 688.
Type:  
  BUG
Influences：  
  check heartbeat status when dble select dbInstance


wenyh1 (approved):
ok

PR #2406, opened 2021-01-08
inner-744-supplement: check argument lack for all native fun

dcy10000:

Reason:  
  BUG inner-744
Type:  
  BUG
Influences：  
   all native fun

### DIFF
--- a/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/query/impl/join/JoinHandler.java
+++ b/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/query/impl/join/JoinHandler.java
@@ -7,7 +7,6 @@ package com.actiontech.dble.backend.mysql.nio.handler.query.impl.join;
 
 
 import com.actiontech.dble.backend.mysql.CharsetUtil;
-
 import com.actiontech.dble.backend.mysql.nio.handler.query.DMLResponseHandler;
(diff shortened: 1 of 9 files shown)


src/main/java/com/actiontech/dble/backend/mysql/nio/handler/query/impl/join/JoinHandler.java
```diff
@@ -231,6 +231,10 @@ protected void ownThreadJob(Object... objects) {
             }
             session.setHandlerEnd(this);
             nextHandler.rowEofResponse(null, isLeft, service);
+        } catch (MySQLOutPutException e) {
```
wenyh1, line 234: Whether consider scenarios where functions are used after statements such as ORDER BY?

dcy10000: I tried some different cases, but can't reappear problems.
Those cases all calculate fun() in MySQL.

dcy10000: I tried a few different situations, but can't reappear problems.
Because in those situations, functions are calculated by MySQL, except for "join".
 Even so，add this to all appropriate handler is a good idea

wenyh1 (approved):
lgtm

Issue #2199, opened 2020-10-22
and ... or statement running error


LUAgam:
* **dble version：2.19.07.0**  
* **preconditions ：**  
no
* **configs：** 

**schema.xml**  

```


```

**rule.xml**  

```


```

**server.xml**  

**only this one table**
```
<table name="model_searchlib" dataNode="dn1,dn2" rule="rule_common_hash"/>

```

* **steps：**    
    step1.
```
CREATE TABLE `model_searchlib` (
  `cc` varchar(255) DEFAULT NULL,
  `aa` varchar(255) DEFAULT NULL,
  `bb` varchar(255) DEFAULT NULL,
  `isshare` int(255) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```
    step2.  
```
select   *  from model_searchlib a
inner join  model_searchlib b on a.cc = b.cc
where  a.aa  ='cn' and b.bb ='jp'
 and (b.isshare=1 or a.isshare=1 )
 LIMIT 0,7
```
* **expect result：**  
    1.Normal return result
* **real result：**  
    1.
```
ERROR 1003 (HY000): Unknown column 'b.isshare' in 'where clause'
```
* **supplements：**  
    1.no

yanhuqing666:
looks fixed in new version

LUAgam:
fixed in new version, associated item #2042

LUAgam:
Misjudgment. The basis for misjudgment is that there are multiple shardingTables under the same schema, which are optimized as er tables at this time. If there is only one shardingTable under a schema, an error will be reported

LUAgam:
influences：
#2073
inner-590

quexiuping:
verified on version:5.7.21-dble-9.9.9.9-9b37a546a8e46e362a9b40b127f135625695ab8f-20210308051611

PR #2407, opened 2021-01-08
fix: when parsing where conditions, you need to match according to table name+table alias

LUAgam:
Reason:  
  BUG #2199 
Type:  
  BUG
Influences：  
when parsing where conditions, you need to match according to table name+table alias
deduplication where filter 
#2073 
inner-590


wenyh1 (approved):
lgtm

PR #2413, opened 2021-01-12
fix: when connecting, the client_found_rows value of the client and dble are inconsistent, the log will prompt

LUAgam:
Reason:  
  Improve  inner-814
Type:  
  Improve  
Influences：  
   fix: when connecting, the client_found_rows value of the client and dble are inconsistent, the log will prompt


PanternBao (approved):
lgtm

PR #2416, opened 2021-01-13
fix: temporary configuration is not rolled back

LUAgam:
Reason:  
  BUG inner-841
Type:  
  BUG
Influences：  
fix: temporary configuration is not rolled back
modify the prompt message


wenyh1 (approved):
lgtm

PR #2417, opened 2021-01-14
Separate load data handler and normal select handler

PanternBao:
Reason:  
  BUG.
Type:  
  BUG
Influences：  
  Separate load data handler and normal select handler


LUAgam (approved):
lgtm

PR #2422, opened 2021-01-19
inner-451: fixed: new node doesn't pause when the cluster is pausing.

dcy10000:

Reason:  
  BUG inner-451
Type:  
  BUG
Influences：  
   1.bootstrap
        2.pause logic


PanternBao (approved):
ok

PR #2424, opened 2021-01-21
fix: adjust the xml verification method from dtd to xsd

LUAgam:
Reason:  
  BUG inner-843/844
Type:  
  BUG 
Influences：  
fix: adjust the xml verification method from dtd to xsd
sequence json is sorted by file order


LUAgam:
need change

wenyh1 (approved):
lgtm

PR #2427, opened 2021-01-21
fix: when connecting, the client_found_rows value of the client and dble are inconsistent, the log will prompt(cherry pick from master #2413)

LUAgam:
Reason:
  Improve inner-814
Type:
  Improve
Influences：
   fix: when connecting, the client_found_rows value of the client and dble are inconsistent, the log will prompt
#2413


PanternBao (approved):
ok

PR #2430, opened 2021-01-22
fix general filePath, and btrace（inner 862&859&858）

wenyh1:
Reason:  
  BUG inner 862、859、858
Type:  
  BUG
Influences：  
   fix general filePath, and btrace


PanternBao (approved):
lgtm

PR #2431, opened 2021-01-22
add batch loaddata function inner 452

ylinzhu:
Reason:  
   Improve 
Type:  
  Improve  
Influences：  
  add batch loaddata function inner 452


src/main/java/com/actiontech/dble/config/model/SystemConfig.java
```diff
@@ -1333,6 +1341,8 @@ public String toString() {
                 ", enableAlert=" + enableAlert +
                 ", maxCharsPerColumn=" + maxCharsPerColumn +
                 ", maxRowSizeToFile=" + maxRowSizeToFile +
+                ", enableBatchLoadData" + enableBatchLoadData +
+                ",enableBatchLoadData=" + enableBatchLoadData +
```
wenyh1, line 1345: repeat

src/main/java/com/actiontech/dble/route/RouteResultsetNode.java
```diff
@@ -29,6 +32,7 @@
     private Boolean runOnSlave = null;
     private AtomicLong multiplexNum;
     private boolean isForUpdate = false;
+    private volatile byte flag;
```
wenyh1, line 35: suggest a clear name

PR #2432, opened 2021-01-25
fix：'begin'，the sql that follows commits implicitly

wenyh1:
Reason:  
  BUG inner 867
Type:  
  BUG
Influences：  
   fix 'begin'，the sql that follows commits implicitly


PanternBao (approved):
lgtm

PR #2433, opened 2021-01-25
fix that dble in performance mode doesn't work and cast exception when show connection

PanternBao:
Reason:  
  BUG #inner 852 and inner 849. 
Type:  
  BUG
Influences：  
  fix that dble in performance mode doesn't work and cast exception when show connection


wenyh1 (approved):
lgtm

PR #2436, opened 2021-01-26
inner-872: fix ArrayOutOfIndex when load data

PanternBao:
Reason:  
  BUG #inner-872.
Type:  
  BUG
Influences：  
    fix ArrayOutOfIndex when load data


wenyh1 (approved):
lgtm

PR #2437, opened 2021-01-26
fix  support set session transaction isolation level REPEATABLE READ  inner 863

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix  support set session transaction isolation level REPEATABLE READ  inner 863


wenyh1 (approved):
ok

PR #2439, opened 2021-01-27
fix add rwSplitUser  need check corresponding dbgroup inner 668 and 667

ylinzhu:
Reason:  
  BUG  
Type:  
  BUG
Influences：  
  fix add rwSplitUser  need check corresponding dbgroup inner 668 and 667


PanternBao (approved):
lgtm

PR #2440, opened 2021-01-27
fix: synchronously modify the value of dbGroup_status in the configuration center during reload

LUAgam:
Reason:  
  BUG inner-878
Type:  
  BUG
Influences：  
fix: synchronously modify the value of dbGroup_status in the configuration center during reload
sequence persistence increases judgment


wenyh1 (approved):
lgtm

PR #2441, opened 2021-01-27
update error message that druid have sql syntax error  inner  861

ylinzhu:
Reason:  
  Improve  
Type:  
    Improve  
Influences：  
  update error message that druid have sql syntax error  inner  861


LUAgam (approved):
ok

PR #2442, opened 2021-01-27
fix: property value is String

LUAgam:
Reason:  
  BUG 
Type:  
  BUG 
Influences：  
  fix: property value is String


PanternBao (approved):
ok

PR #2443, opened 2021-01-27
fix that dble hang when client keeps compression on

PanternBao:
Reason:  
  BUG.
Type:  
  BUG
Influences：  
  fix that dble hang when client keeps compression on


wenyh1 (approved):
lgtm

PR #2446, opened 2021-01-28
fix: errors caused by xsd

LUAgam:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix: errors caused by xsd


LUAgam:
Organize submission records

PR #2447, opened 2021-01-29
fix: errors caused by xsd

LUAgam:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
fix: errors caused by xsd

fix: pure read-write separation scenario without loading sharding.xml


PanternBao (approved):
lgtm

PR #2449, opened 2021-01-29
update camel case and units of number inner 517

ylinzhu:
Reason:  
   Improve
Type:  
  Improve  
Influences：  
  update camel case and units of number inner 517


LUAgam (approved):
ok

PR #2450, opened 2021-01-29
fix that multi statement doesn't work

PanternBao:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix that multi statement doesn't work


LUAgam (approved):
lgtm

PR #2451, opened 2021-01-31
fix add rwsplit mode when  all  slave node is alive  that request to master node inner 725

ylinzhu:
Reason:  
  Improve 
Type:  
  Improve  
Influences：  
 fix add rwsplit mode when  all  slave node is alive  that request to master node inner 725

### DIFF
--- a/src/main/java/com/actiontech/dble/config/model/db/DbGroupConfig.java
+++ b/src/main/java/com/actiontech/dble/config/model/db/DbGroupConfig.java
@@ -70,7 +70,7 @@ public class DbGroupConfig {
         if (rwSplitMode >= 0 && rwSplitMode <= 3) {
             this.rwSplitMode = rwSplitMode;
         } else {
-            throw new ConfigException("dbGroup " + name + " rwSplitMode should be between 0 and 2!");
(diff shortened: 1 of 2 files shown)


src/main/java/com/actiontech/dble/config/model/db/DbGroupConfig.java
```diff
@@ -67,7 +67,7 @@ public int getRwSplitMode() {
     }
 
     public void setRwSplitMode(int rwSplitMode) {
-        if (rwSplitMode >= 0 && rwSplitMode <= 2) {
+        if (rwSplitMode >= 0 && rwSplitMode <= 3) {
             this.rwSplitMode = rwSplitMode;
         } else {
             throw new ConfigException("dbGroup " + name + " rwSplitMode should be between 0 and 2!");
```
yanhuqing666, line 73: between 0 and ?

ylinzhu: between 0 and 3

PanternBao:
may need to think of the scenario that user forces slave by hint.

ylinzhu:
year, really not considered, has been modified

PanternBao (approved):
lgtm

PR #2453, opened 2021-01-31
fix:  prepare sql use single frontConnection not multiple frontConnection inner 594

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
    conflicts:src/main/java/com/actiontech/dble/services/rwsplit/RWSplitQueryHandler.java
    fix:  prepare sql use single frontConnection not multiple frontConnection inner 594

### DIFF
--- a/src/main/java/com/actiontech/dble/rwsplit/RWSplitNonBlockingSession.java
+++ b/src/main/java/com/actiontech/dble/rwsplit/RWSplitNonBlockingSession.java
@@ -4,17 +4,25 @@ import com.actiontech.dble.backend.datasource.PhysicalDbGroup;
 import com.actiontech.dble.backend.datasource.PhysicalDbInstance;
 import com.actiontech.dble.config.ErrorCode;
 import com.actiontech.dble.net.connection.BackendConnection;
+import com.actiontech.dble.plan.common.exception.MySQLOutPutException;
 import com.actiontech.dble.services.rwsplit.Callback;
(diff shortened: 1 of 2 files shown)


src/main/java/com/actiontech/dble/rwsplit/RWSplitNonBlockingSession.java
```diff
@@ -32,6 +41,28 @@ public void execute(Boolean master, Callback callback) {
         execute(master, null, callback);
     }
 
+    public void execute(Boolean master, Callback callback, String sql) {
+        try {
+            SQLStatement statement = parseSQL(sql);
+            if (statement instanceof MySqlPrepareStatement) {
+                String simpleName = ((MySqlPrepareStatement) statement).getName().getSimpleName();
+                nameSet.add(simpleName);
+                rwSplitService.setTxStart(true);
```
wenyh1, line 50: application point？
prepare has no practical relationship with tx; (doing so may lead to other aspects of influence)
If it's for some purpose, try something else.

LUAgam:
Does [DROP PREPARE](https://dev.mysql.com/doc/refman/8.0/en/deallocate-prepare.html) need to be implemented?

ylinzhu:
druid not support  DROP PREPARE

wenyh1 (dismissed):
wait for the follow-up support DROP PREPARE, need review.

PR #2456, opened 2021-02-02
feat: optimize split

LUAgam:
Reason:
  Improve optimize split inner-790
Type:
  Improve
Influences：
  optimize split


wenyh1 (approved):
lgtm

PR #2457, opened 2021-02-02
inner-882&879: add config check for cursor settings

dcy10000:

Reason:  
  BUG inner-882&879
Type:  
  BUG
Influences：  
   config load


LUAgam (approved):
ok

PR #2458, opened 2021-02-02
inner-851-bugfix: select with no from cause error in mysql5.7

dcy10000:

Reason:  
  BUG inner-851-bugfix
Type:  
  BUG
Influences：  
   prepare


wenyh1 (approved):
lgtm

PR #2459, opened 2021-02-02
inner-884: prepare phase cause NullPointException

dcy10000:

Reason:  
  BUG inner-884
Type:  
  BUG
Influences：  
   prepare


LUAgam (approved):
ok

PR #2460, opened 2021-02-02
inner-875: report error if run different PAUSE command

dcy10000:

Reason:  
  BUG inner-875
Type:  
  BUG
Influences：  
   pause


LUAgam (approved):
lgtm

PR #2462, opened 2021-02-02
fix：execute exit/quit command，general log has problem that will not be logged; And headers are logged twice（inner 865&891）

wenyh1:
Reason:  
  BUG inner 865、891
Type:  
  BUG/Improve  
Influences：  
   fix execute exit/quit command，general log has problem that will not be logged; And headers are logged twice


PanternBao (approved):
lgtm

PR #2463, opened 2021-02-02
fix multiple users are need cleaned up sysVariables and usrVariables inner 880

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix multiple users are need cleaned up sysVariables and usrVariables inner 880


yanhuqing666 (approved):
ok

PR #2464, opened 2021-02-02
fix: properties read and write order

LUAgam:
Reason:  
  BUG innert-843/883
Type:  
  BUG
Influences：  
fix: properties read and write order
fix: connection service is null


PanternBao (approved):
lgtm

PR #2465, opened 2021-02-02
inner-898: stop node when pausing cause error.

dcy10000:

Reason:  
  BUG inner-898
Type:  
  BUG
Influences：  
   pause for zk


PanternBao (approved):
lgtm

PR #2466, opened 2021-02-03
fix: auth plugin is empty when Navicat 11 is connected

LUAgam:
Reason:  
  BUG #2389 
Type:  
  BUG
Influences：  
fix: auth plugin is empty when Navicat 11 is connected


wenyh1 (approved):
lgtm

PR #2467, opened 2021-02-03
load data send to mysql  or receive from mysql, update the readTime of the frontend connection inner 418

ylinzhu:
Reason:  
  BUG  and Improve  
Type:  
  BUG and Improve  
Influences：  
  load data send to mysql  or receive from mysql, update the readTime of the frontend connection inner 418.
       fix the parameters resulting from refactoring are not available.


LUAgam (approved):
lgtm

PR #2468, opened 2021-02-03
fix dryrun check database and then check table again inner 672

ylinzhu:
Reason:  
  BUG  
Type:  
  BUG
Influences：  
  fix dryrun check database and then check table again inner 672


PanternBao (approved):
lgtm

PR #2469, opened 2021-02-03
fix that set statement that contains `.` doesn't work, eg: set @.abc_123$=8

PanternBao:
Reason:  
  BUG #inner-889. 
Type:  
  BUG
Influences：  
  fix that set statement that contains `.` doesn't work, eg: set @.abc_123$=8


wenyh1 (approved):
ok

PR #2471, opened 2021-02-05
fix bin function error inner 746

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix bin function error inner 746.
       compatibility was done in the last step


src/main/java/com/actiontech/dble/plan/common/item/function/mathsfunc/ItemFuncConv.java
```diff
@@ -25,6 +26,9 @@ public final String funcName() {
 
     @Override
     public String valStr() {
+        if (args.size() == 1) {
```
LUAgam, line 29: does it need to be handled separately BIN()?

ylinzhu: doc   :    https://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_bin

content ： Returns a string representation of the binary value of N, where N is a longlong (BIGINT) number. This is equivalent to CONV(N,10,2). Returns NULL if N is NULL.

LUAgam (approved):
ok

PR #2472, opened 2021-02-05
fix: null judgment and pack tar.gz to remove xsd files

LUAgam:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
fix: null judgment and pack tar.gz to remove xsd files


PanternBao (approved):
lgtm

PR #2474, opened 2021-02-05
inner-884-bugfix: prepare phase cause NullPointException

dcy10000:

Reason:  
  BUG inner-884-bugfix
Type:  
  BUG
Influences：


dcy10000:
not avaliable yet

PR #2475, opened 2021-02-07
fix: hint statement use executeSql

LUAgam:
Reason:  
  BUG inner-902/906
Type:  
  BUG
Influences：  
fix: hint statement use executeSql
fix: wrong judgment when parsing reload


PanternBao (approved):
lgtm

PR #2476, opened 2021-02-07
inner-884-bugfix: prepare phase cause error

dcy10000:

Reason:  
  BUG inner-884
Type:  
  BUG
Influences：  
   prepare


PanternBao (approved):
lgtm

PR #2481, opened 2021-02-22
fix lost records for CLOSE_STMT packet in general Log(inner-865)

wenyh1:
Reason:  
  BUG inner  865
Type:  
  BUG
Influences：  
   fix lost records for CLOSE_STMT packet in general Log


LUAgam (approved):
ok

PR #2482, opened 2021-02-22
fix:  prepare sql use single frontConnection not multiple frontConnection inner 594

ylinzhu:
Reason:
  BUG
Type:
  BUG
Influences：
conflicts:src/main/java/com/actiontech/dble/services/rwsplit/RWSplitQueryHandler.java
fix: prepare sql use single frontConnection not multiple frontConnection inner 594


ylinzhu:
mysql support sql like {DEALLOCATE | DROP} PREPARE stmt_name. Now  only  support sql like  DEALLOCATE  PREPARE  stmt_name , because druid don't support sql like DROP PREPARE stmt_name .

wenyh1 (approved):
ok

PR #2486, opened 2021-02-23
inner-913:add state column for dble backend connections

PanternBao:
Reason:  
   Improve #inner 913.  
Type:  
   Improve  
Influences：  
   add state column for dble backend connections


ylinzhu:
lgtm

ylinzhu (approved):
ok

PR #2487, opened 2021-02-23
inner-916:props of testOn* take no effect

PanternBao:
Reason:  
  BUG #inner 916.
Type:  
  BUG
Influences：  
  props of testOn* take no effect


LUAgam (approved):
ok

PR #2490, opened 2021-02-23
inner-879-bugfix2: in config init, string to Integer report error.but  dble start successfully.

dcy10000:

Reason:  
  BUG inner-879-bugfix2
Type:  
  BUG
Influences：  
  config init


PanternBao (approved):
lgtm

PR #2491, opened 2021-02-23
inner-910: pause operation shouldn't resume when caller is down

dcy10000:

Reason:  
  BUG inner-910
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
lgtm

PR #2494, opened 2021-02-24
support day function inner 663

ylinzhu:
Reason:  
  Improve 
Type:  
  Improve  
Influences：  
  support day function inner 663


LUAgam (approved):
ok

PR #2495, opened 2021-02-24
fix: no need to synchronize dbGroupStatus before reload operation

LUAgam:
Reason:  
  BUG inner-918/922
Type:  
  BUG
Influences：  
fix: no need to synchronize dbGroupStatus before reload operation(inner-922)
fix: the user order in user.xml does not need to be specified(inner-918)


wenyh1 (approved):
lgtm

PR #2498, opened 2021-02-24
fix SUBSTRING function inner 757

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG 
Influences：  
  fix SUBSTRING function inner 757


wenyh1 (approved):
ok

PR #2500, opened 2021-02-25
fix:sql statistic's small bugs

wenyh1:
Reason:  
  BUG inner 923\926\927\928\931\932\934
Type:  
  BUG
Influences：  
   fix:sql statistic's small bugs


dcy10000 (approved):
lgtm

PR #2504, opened 2021-03-02
fix: when the schemas of the front and back ends are inconsistent, reset the connection status

LUAgam:
Reason:  
  BUG inner-943
Type:  
  BUG
Influences：  
  fix: when the schemas of the front and back ends are inconsistent, reset the connection status


LUAgam:
need to be updated

wenyh1 (approved):
lgtm

Issue #2377, opened 2020-12-15
Can't use both  show @@processlist and RwSplitService


dcy10000:
* **dble version：**  master
* **preconditions ：**  
no
* **configs：** 

* **steps：**    
    step1. login in to dble using any rwSplitUser.
    step2. run 'show @@processlist' in 9066 port
* **expect result：**  
    show all connections in dble.
* **real result：**  
get error.Results are showed as follows.
```
mysql> show @@processlist;
ERROR 1003 (HY000): get error call manager command com.actiontech.dble.services.rwsplit.RWSplitService cannot be cast to com.actiontech.dble.services.mysqlsharding.ShardingService

```
* **supplements：**  
    1.
![image](https://user-images.githubusercontent.com/26567055/102195006-fb11ee80-3ef8-11eb-869f-f0a9ca9b0704.png)

PanternBao:
change "shardingNode" column to "db_instance" column

quexiuping:
Need to modify the document @PanternBao

quexiuping:
verified on：5.7.21-dble-9.9.9.9-c05135ec7239c645112df117738a3ede2a5a1321-20210304083333

PR #2505, opened 2021-03-02
issue 2377 - support rw connection in show @@processlist

PanternBao:
Reason:  
  BUG #2377.
Type:  
  BUG
Influences：  
  support rw connection in show @@processlist


wenyh1 (approved):
lgtm

PR #2508, opened 2021-03-03
field list need synchronize context

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  field list need synchronize context


dcy10000 (approved):
lgtm

PR #2509, opened 2021-03-03
inner 816 - when rwSplitMode = 0 and master of mysql group is changed, dble need to build connection pool  for salve.

PanternBao:
Reason:  
  BUG inner 816
Type:  
  BUG
Influences：  
   when rwSplitMode = 0 and master of mysql group is changed, dble need to build connection pool for salve.


wenyh1 (approved):
lgtm

PR #2511, opened 2021-03-04
fix dryrun  multiple nodes have incorrect data inner 949

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG  
Influences：  
  fix dryrun multiple nodes have incorrect data inner 949


wenyh1 (approved):
lgtm

PR #2512, opened 2021-03-04
Inner 804&953 - fakeMysqlVersion bug and select for  share support

dcy10000:
Reason:  
  BUG 804&953
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
lgtm

PR #2513, opened 2021-03-04
correct 'primary' column in dble_db_instance manager table

PanternBao:
Reason:  
  BUG inner-954
Type:  
  BUG  
Influences：  
  correct 'primary' column in dble_db_instance manager table


dcy10000 (approved):
lgtm

PR #2516, opened 2021-03-05
modify  log4j pattern for developer

dcy10000:

Reason:  
  Provide convenience for development。
       A pattern like this will make the log can be clickable every line.

Type:  
  BUG
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #2519, opened 2021-03-09
testOnBorrows takes no effect when backend connection was take directly from connection pool

PanternBao:
Reason:  
  BUG #inner 915. 
Type:  
  BUG
Influences：  
  testOnBorrows takes no effect when backend connection was take directly from connection pool


wenyh1 (approved):
ok

PR #2521, opened 2021-03-10
Result of show @@reload_status is empty instead of default row when there is no reload action.

PanternBao:
Reason:  
     Improve #inner 952.  
Type:  
  Improve  
Influences：  
  Result of show @@reload_status is empty instead of default row when there is no reload action.


wenyh1 (approved):
ok

PR #2523, opened 2021-03-11
fix connection crash when load data inner 828

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix connection crash when load data inner 828


dcy10000 (approved):
lgtm

PR #2524, opened 2021-03-11
fix conv function format error return value  innner 908

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix connection crash when load data inner 828


dcy10000 (approved):
lgtm

dcy10000 (approved):
lgtm

PR #2525, opened 2021-03-11
fix SUBSTRING function  index out of range inner 945

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix SUBSTRING function  index out of range inner 945


dcy10000 (approved):
ok

PR #2526, opened 2021-03-12
inner 912 supplement - ddl response skip init event

dcy10000:

Reason:  
    inner 912 supplement
Type:  
  BUG
Influences：  
   ddl


PanternBao (approved):
lgtm

PR #2527, opened 2021-03-12
Result of dble_reload_status manager table is empty instead of default row when there is no reload action.

PanternBao:
Reason:  
  Improve #inner 965.  
Type:  
  Improve  
Influences：  
  Result of dble_reload_status manager table is empty instead of default row when there is no reload action.


dcy10000 (approved):
lgtm

PR #2528, opened 2021-03-12
[inner 874] fixed npe when kill xa query

dcy10000:

Reason:  
  BUG inner 874
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
lgtm

PR #2531, opened 2021-03-12
fix dble crash when load data statement have local field inner 966

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix dble crash when load data statement have local field inner 966


LUAgam (approved):
ok

PR #2532, opened 2021-03-12
the log shows an uncaught exception inner 967

ylinzhu:
Reason:  
  Improve 
Type:  
  Improve  
Influences：  
  fix dble crash when load data statement have local field inner 966


wenyh1 (approved):
ok

PR #2535, opened 2021-03-16
inner 915 - props of testOn* take no effect

PanternBao:
Reason:  
   Improve #inner 915.  
Type:  
   Improve  
Influences：  
   props of testOn* take no effect


wenyh1 (approved):
ok

PR #2536, opened 2021-03-16
fix: show @@help increase thread parameters

LUAgam:
Reason:  
  BUG inner-876
Type:  
  BUG
Influences：  
show @@ help increase thread parameters


wenyh1 (approved):
ok

PR #2537, opened 2021-03-16
fix NullPointException using CURRENT_TIME function

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
   eg: select id from  tb_hash_sharding union SELECT CURRENT_TIME /*allow_diff*/;


wenyh1 (approved):
lgtm

PR #2539, opened 2021-03-17
fix:'lock tables ...' sql ,tx_count does not add up

wenyh1:
Reason:  
  BUG inner 959
Type:  
  BUG
Influences：  
   fix:'lock tables ...' sql ,tx_count does not add up


ylinzhu (approved):
lgtm

PR #2540, opened 2021-03-17
fix: db.xml allows the content to be empty

LUAgam:
Reason:  
  BUG inner-976
Type:  
  BUG
Influences：  
fix: db.xml allows the content to be empty


PanternBao (approved):
lgtm

PR #2542, opened 2021-03-17
fix load data prompt optimized

ylinzhu:
Reason:  
  Improve   
Type:  
   Improve  
Influences：  
   fix bug inner 961 and 962  and 963


wenyh1 (approved):
ok

PR #2547, opened 2021-03-22
ok packet of conn heartbeat returns when heartTimeout isn't initialized

PanternBao:
Reason:  
  BUG #inner 987.
Type:  
  BUG
Influences：  
  ok packet of conn heartbeat returns when heartTimeout isn't initialized


LUAgam (approved):
lgtm

PR #2550, opened 2021-03-23
Sharding user configuring the blacklist fails to execute begin

wenyh1:
Reason:  
  BUG inner 944
Type:  
  BUG
Influences：  
   fix :Sharding user configuring the blacklist fails to execute begin


ylinzhu (approved):
lgtm

PR #2551, opened 2021-03-24
fix: eliminate invalid connections

LUAgam:
Reason:  
  BUG inner-942
Type:  
  BUG
Influences：  
fix: eliminate invalid connections


PanternBao (approved):
lgtm

PR #2552, opened 2021-03-24
fix load data error inner 1000

ylinzhu:
Reason:  
  BUG  
Type:  
  BUG  
Influences：  
  fix load data error inner 1000 and 986


LUAgam (approved):
ok

PR #2554, opened 2021-03-24
fix 'begin' to repeat on sql_log table

wenyh1:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  fix 'begin' to repeat on sql_log table


PanternBao (approved):
lgtm

PR #2562, opened 2021-03-26
fix: Zookeeper 3.4.13 or higher version causes startup error

LUAgam:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
fix: Zookeeper 3.4.13 or higher version causes startup error


dcy10000 (approved):
lgtm

PR #2564, opened 2021-03-26
fix skip idleTimeout checks inner 418 and  load data error when mysql crash

ylinzhu:
Reason:  
  BUG & Improve
Type:  
  BUG & Improve  
Influences：  
   fix skip idleTimeout checks inner 418 and  load data error when mysql crash


PanternBao (approved):
lgtm

PR #2567, opened 2021-03-30
fix: the listening thread(ddl/view/ha) is not started(cherry-pick 2559)

LUAgam:
Reason:
  BUG inner-1019
Type:
  BUG
Influences：
fix: the listening thread(ddl/view/ha) is not started


ylinzhu (approved):
lgtm

PR #2569, opened 2021-03-30
inner-914-supplement：limit size for frontQueue

dcy10000:

Reason:  
  inner-914-supplement
Type:  
  Improve  
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #2572, opened 2021-04-01
fix bug about sampling

PanternBao:
Reason:  
  BUG #inner 1032, inner 1033.
Type:  
  BUG
Influences：  
  fix bug about sampling


LUAgam (approved):
ok

PR #2573, opened 2021-04-01
fix in some cases there is no update time

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG  
Influences：  
   fix in some cases there is no update time inner 418


ylinzhu:
Travis CI is not detected, so refire

dcy10000 (approved):
ok

PR #2574, opened 2021-04-01
fix: xa retry mechanism update

LUAgam:
Reason:  
  BUG inner-DMP-7363
Type:  
  BUG
Influences：  
fix: xa retry mechanism update


PanternBao (approved):
lgtm

PR #2576, opened 2021-04-08
inner-1046: fixed: init view from empty file

dcy10000:

Reason:  
  BUG inner-1046
Type:  
  BUG
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #2577, opened 2021-04-08
packetId is incorrect when doing ddl

PanternBao:
Reason:  
  BUG #inner 1044. 
Type:  
  BUG
Influences：  
  packetId is incorrect when doing ddl


wenyh1 (approved):
lgtm

PR #2579, opened 2021-04-08
fix flow control error when turn off flow control inner 418

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix flow control error when turn off flow control inner 418


dcy10000 (approved):
lgtm

PR #2582, opened 2021-04-09
avoid npe when keeping useSerializableMode on

PanternBao:
Reason:  
  BUG #inner 1047. 
Type:  
  BUG
Influences：  
  avoid npe when keeping useSerializableMode on


wenyh1 (approved):
ok

wenyh1 (approved):
ok

PR #2583, opened 2021-04-09
dble hang when backend connection is closed in process of testOnCreate

PanternBao:
Reason:  
  BUG #inner 1028. 
Type:  
  BUG
Influences：  
  dble hang when backend connection is closed in process of testOnCreate


dcy10000 (approved):
ok

PR #2584, opened 2021-04-12
reloading samplingRate doesn't work

PanternBao:
Reason:  
  BUG #inner 1048.
Type:  
  BUG 
Influences：  
  reloading samplingRate doesn't work


dcy10000 (approved):
ok

PR #2585, opened 2021-04-13
fix in some cases there is no update time(cherry-pick 2579 and 2573)

ylinzhu:
Reason:  
  BUG  
Type:  
  BUG
Influences：  
   fix in some cases there is no update time


PanternBao (approved):
ok

PR #2586, opened 2021-04-13
fix: when dbGroup is not configured, it should start successfully

LUAgam:
Reason:  
  BUG inner-1043
Type:  
  BUG
Influences：  
fix: when dbGroup is not configured, it should start successfully
DBLE0REQ-1043


dcy10000 (approved):
ok

PR #2589, opened 2021-04-14
fix：table name was not resolved in the hint sql

wenyh1:
Reason:  
  BUG inner 1051
Type:  
  BUG
Influences：  
  fix：table name was not resolved in the hint sql


PanternBao (approved):
lgtm

PR #2590, opened 2021-04-14
fix: after configuring blacklist, execute error sql, prompt is not reasonable

wenyh1:
Reason:  
  BUG inner 1052
Type:  
  BUG
Influences：  
   fix: after configuring blacklist, execute error sql, prompt is not reasonable


LUAgam (approved):
ok

PR #2591, opened 2021-04-14
fix: table name was not resolved in the hint sql(cherry pick from 3.21.02.99)

wenyh1:
Reason:  
  BUG inner 1051
Type:  
  BUG
Influences：  
   fix: table name was not resolved in the hint sql(cherry pick from 3.21.02.99)


PanternBao (approved):
ok

PR #2599, opened 2021-04-19
fix some bug from 3.21.02.0

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  cherrry pick from 3.21.02.99


dcy10000 (approved):
ok

PR #2605, opened 2021-04-20
build arg of Dockerfile can't be assigned

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  build arg of Dockerfile can't be assigned


LUAgam (approved):
ok

PR #2607, opened 2021-04-21
support session read only

PanternBao:
Reason:  
  Improve #inner 758.  
Type:  
  Improve  
Influences：  
  support session read only


ylinzhu (approved):
ok

PR #2608, opened 2021-04-21
the sequence is params,field in com_stmt_prepare response

PanternBao:
Reason:  
  BUG
Type:  
  BUG
Influences：  
 the sequence is params,field in com_stmt_prepare response


dcy10000 (approved):
lgtm

PR #2612, opened 2021-04-25
inner-1064 in subquery explain

yanhuqing666:
Reason:  
  BUG inner-1064
Type:  
  BUG 
Influences：  
   in subquery explain


ylinzhu (approved):
ok

PR #2614, opened 2021-04-26
fix begin sql not set autocommit field to false inner 801

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG 
Influences：  
  fix begin sql not set autocommit field to false inner 801

### DIFF
--- a/src/main/java/com/actiontech/dble/server/handler/BeginHandler.java
+++ b/src/main/java/com/actiontech/dble/server/handler/BeginHandler.java
@@ -20,6 +20,7 @@ public final class BeginHandler {
             service.beginInTx(stmt);
         } else {
             service.setTxStart(true);
+            service.setAutocommit(false);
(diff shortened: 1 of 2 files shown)


src/main/java/com/actiontech/dble/server/handler/BeginHandler.java
```diff
@@ -20,6 +20,7 @@ public static void handle(String stmt, ShardingService service) {
             service.beginInTx(stmt);
         } else {
             service.setTxStart(true);
+            service.setAutocommit(false);
```
LUAgam, line 23: txStart and autocommit have different meanings

yanhuqing666: +1

LUAgam (approved):
ok

PR #2618, opened 2021-04-28
inner 314 - remove show @@syslog

PanternBao:
Reason:  
  Improve #inner 314.  
Type:  
  Improve  
Influences：  
  remove show @@syslog and log @@[file=logname limit=numberOfRow key=keyword regex=regex]


ylinzhu (approved):
ok

PR #2619, opened 2021-04-29
inner-690: fix AsynchronousCloseException while read

dcy10000:

Reason:  
  BUG inner-690
Type:  
  BUG
Influences：  
   ioReactor


wenyh1 (approved):
lgtm

PR #2621, opened 2021-04-30
Optimizing Transactions statistics

wenyh1:
Reason:  
  BUG inner 982
Type:  
  BUG
Influences：  
   Optimizing Transactions statistics


ylinzhu (approved):
ok

PR #2624, opened 2021-04-30
fix bug that "set autocommit=1" doesn't work,eg set autocommit=0;set autocommit=1;dml;rollback; dml is rollback

PanternBao:
Reason:  
  BUG #inner1074.
Type:  
  BUG
Influences：  
  set autocommit=0;set autocommit=1;dml;rollback; dml is rollback


wenyh1 (approved):
lgtm

PR #2627, opened 2021-05-07
enlarge metaspace and shrink thread stack size

PanternBao:
Reason:  
  Improve.  
Type:  
  Improve  
Influences：  
  enlarge metaspace and shrink thread stack size


dcy10000 (approved):
ok

PR #2629, opened 2021-05-07
optimize load data  information inner 856

ylinzhu:
Reason:  
   Improve 
Type:  
  Improve  
Influences：  
  optimize load data  information inner 856


PanternBao (approved):
lgtm

PR #2632, opened 2021-05-11
inner-1006-bugfix:fix cursor for blob type

dcy10000:

Reason:  
  BUG inner-1006-bugfix
Type:  
  BUG
Influences：  
 fix cursor for blob type


LUAgam (approved):
ok

PR #2633, opened 2021-05-11
fix bug about sql digest

PanternBao:
Reason:  
  BUG #inner 1087,1088,1093,1094.
Type:  
  BUG
Influences：  
  fix bug about sql digest


wenyh1 (approved):
lgtm

PR #2635, opened 2021-05-11
fix table name  may not correct inner 905

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix table name  may not correct inner 905


wenyh1 (approved):
lgtm

Issue #2638, opened 2021-05-12
Keep slow log on, dble may occur oom


PanternBao:
* **dble version：**  3.21.02.0 or earily
* **preconditions ：**  
```
drop table if exists sharding_2_t1;
create table sharding_2_t1(id int , name varchar(12));
```
* **configs：** 

**bootstrap.cnf**  

```
-DenableSlowLog=1
```
**db.xml**  

```
<dbGroup name="dbGroup2" rwSplitMode="0" delayThreshold="100">
    <heartbeat>select 1</heartbeat>
    <dbInstance name="instanceM2" url="127.0.0.1:3307" user="root" password="123456" maxCon="10" minCon="5"
                primary="true"/>
</dbGroup>
```

**user.xml**  

```
<shardingUser name="root" password="123456" schemas="schema1" maxCon="20"/>
```

**sharding.xml**  

```
<schema name="schema1" sqlMaxLimit="100">
    <shardingTable name="sharding_2_t1" shardingNode="dn1,dn2" function="hash-two" shardingColumn="id"/>
</schema>

<shardingNode name="dn1" dbGroup="dbGroup2" database="db1"/>
<shardingNode name="dn2" dbGroup="dbGroup2" database="db2"/>

<function name="hash-two" class="Hash">
    <property name="partitionCount">2</property>
    <property name="partitionLength">1</property>
</function>

```

* **steps：**    
    step1.  insert into sharding_2_t1 value(1,'1');
    step2. /\*!dble:shardingNode=dn1\*/select * from sharding_2_t1;  may execute many times
* **expect result：**  
    1.dble run normally
* **real result：**  
    1.dble oom
* **supplements：**  
    1.

quexiuping:
verified on :5.7.21-dble-9.9.9.9-2fba64ab7af7e4fd94dfcdca96a06b09737d0a0d-20210712061820

PR #2640, opened 2021-05-12
fix bug that keep slow log on, dble may occur oom

PanternBao:
Reason:  
  BUG #2638 .
Type:  
  BUG
Influences：  
  fix bug that keep slow log on, dble may occur oom


LUAguoaomen426 (approved):
lgtm

LUAgam (approved):
lgtm

PR #2643, opened 2021-05-13
add log to observe instead of oom for issue 2638

PanternBao:
Reason:  
  Improve #2638.
Type:  
  Improve  
Influences：  
   add log to observe instead of oom for issue 2638


LUAgam (approved):
ok

PR #2645, opened 2021-05-13
just drop table that do not exist should prompt error

wenyh1:
Reason:  
  BUG
Type:  
  BUG
Influences：  
   just drop table that do not exist should prompt error


ylinzhu (approved):
ok

PR #2649, opened 2021-05-13
inner-1001 - allocate n * chunksize heap byte buffer instead of just good size when direct byte buffer isn't enough

PanternBao:
Reason:  
  BUG #inner 1001.
Type:  
  BUG
Influences：  
  allocate n * chunksize heap byte buffer instead of just good size when direct byte buffer isn't enough


dcy10000 (approved):
ok

PR #2617, opened 2021-04-28
fix: optimize complex queries

LUAgam:
Reason:  
  Improve  #2408/inner-504
Type:  
  Improve  
Influences：  
optimize for the same table in complex queries
when conditions permit, multi-node queries are optimized to single-node queries

Issue #2408, opened 2021-01-11
sharding routing cannot correctly route to the correct sharding database at "left join" and "union all"


prostranger:
* **dble version：**  
* 5.6.29-dble-3.20.07.1-70086407f81463c0d7acefdfd2f4f58fe885d1b9-20201021103611
* **preconditions ：**  
no
* **configs：** 
drop table if exists shard_table_a;
create table shard_table_a(
id bigint not null,
shard_key varchar(4) not null,
filed_a varchar(10),
primary key(id)
);

drop table if exists shard_table_b;
create table shard_table_b(
id bigint not null,
shard_key varchar(4) not null,
filed_b varchar(10),
primary key(id)
);

sharding enum：
partition-enum.txt
0000=0
5104=0
5107=1
5101=2
5108=3
5132=4
5109=5
5103=6
5114=7
7178=7
5118=8
5120=9
5113=10
5100=11
5133=12
5111=13
5117=14
5119=15
5116=16
5115=17
5106=18
5110=19
5134=20
5105=21

**cluster.cnf**  

```


```

**bootstrap.cnf**  

```


```
**db.xml**  

```
<dble:db xmlns:dble="http://dble.cloud/">
    <dbGroup name="Hostname1" rwSplitMode="0" delayThreshold="100">
        <heartbeat>show slave status</heartbeat>
        <dbInstance name="hostM1" url="192.168.0.99:3306" user="test_dc" password="xxx" maxCon="1000" minCon="100" primary="true" usingDecrypt="0">
            <property name="heartbeatPeriodMillis">10000</property>
            <property name="timeBetweenEvictionRunsMillis">300000</property>
            <property name="idleTimeout">1800000</property>
        </dbInstance>
        <dbInstance name="hostR1" url="192.168.0.196:3306" user="test_dc" password="xxx" maxCon="1000" minCon="100" usingDecrypt="0">
            <property name="heartbeatPeriodMillis">10000</property>
            <property name="timeBetweenEvictionRunsMillis">300000</property>
            <property name="idleTimeout">1800000</property>
        </dbInstance>
    </dbGroup>
    <dbGroup name="Hostname2" rwSplitMode="0" delayThreshold="100">
        <heartbeat>show slave status</heartbeat>
        <dbInstance name="hostM1" url="192.168.0.78:3306" user="test_dc" password="xxx" maxCon="1000" minCon="100" primary="true" usingDecrypt="0">
            <property name="heartbeatPeriodMillis">10000</property>
[...]

PR #2651, opened 2021-05-13
fix: remove schema when merging and issuing

LUAgam:
Reason:  
  BUG #2408  inner-504
Type:  
  BUG
Influences：  
fix: remove schema when merging and issuing
after join and condition is spliced to where
#2617


wenyh1 (commented):
lgtm

wenyh1 (approved):
lgtm

PR #2653, opened 2021-05-17
in subquery  not transform join inner 1055

ylinzhu:
Reason:  
   Improve inner 1055
Type:  
  Improve  
Influences：  
   in subquery not transform join inner 1055


LUAgam (approved):
lgtm

PR #2654, opened 2021-05-17
BUG #2638 - fix bug that keep slow log on, dble may occur oom

PanternBao:
Reason:  
  BUG #2638. 
Type:  
  BUG
Influences：  
   fix bug that keep slow log on, dble may occur oom


yanhuqing666 (approved):
ok

PR #2656, opened 2021-05-17
fix add dbGroup check inner 1121

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix add dbGroup check inner 1121


dcy10000 (approved):

dcy10000 (approved):
ok

PR #2657, opened 2021-05-18
fix:complex queries including sub-queries do not need to be optimized for single node delivery

LUAgam:
Reason:  
  BUG inner-1119
Type:  
  BUG
Influences：  
fix:complex queries including sub-queries do not need to be optimized for single node delivery
eg:select count(*) from sharding_4_t1 where id =(select id from sharding_4_t1 where id=1)


dcy10000 (approved):
lgtm

PR #2658, opened 2021-05-18
fix: detect XA residual bugs

wenyh1:
Reason:  
  BUG inner 1120、1123
Type:  
  BUG
Influences：  
   fix detect XA residual bugs


PanternBao (approved):
lgtm

PR #2659, opened 2021-05-18
fix ArrayIndexOutOfBoundException using big packet

PanternBao:
Reason:  
  BUG. 
Type:  
  BUG  
Influences：  
  fix ArrayIndexOutOfBoundException using big packet


dcy10000 (approved):
lgtm

PR #2660, opened 2021-05-18
trace is not supported for rwSplitUser

wenyh1:
Reason:  
  BUG inner 1108
Type:  
  BUG
Influences：  
   trace is not supported for rwSplitUser


dcy10000 (approved):
ok

PR #2662, opened 2021-05-18
inner-1115 : fix cursor hang

dcy10000:

Reason:  
  inner-1115
Type:  
  BUG
Influences：  
  fix cursor hang


ylinzhu (approved):
ok

PR #2665, opened 2021-05-18
fix: capture Throwable to ensure that the thread will not exit when an error or exception occurs

LUAgam:
Reason:  
  BUG inner-957
Type:  
  BUG
Influences：  
fix: capture Throwable to ensure that the thread will not exit when an error or exception occurs


LUAgam:
need to change

ylinzhu (approved):
ok

PR #2666, opened 2021-05-19
fix dbGroup will disabled need more explicit error reporting inner 893

ylinzhu:
Reason:  
   Improve 
Type:  
     Improve  
Influences：  
   fix dbGroup will disabled need more explicit error reporting inner 893 
        fix start dble not check dbgroup


LUAgam (approved):
ok

PR #2667, opened 2021-05-19
fix rwSplitMode check inner 1081

ylinzhu:
Reason:  
  BUG  
Type:  
  BUG
Influences：  
   fix rwSplitMode check inner 1081


dcy10000 (approved):
ok

PR #2669, opened 2021-05-19
remove file @@ manager command

PanternBao:
Reason:  
  Improve #inner 1091.  
Type:  
  Improve  
Influences：  
  remove file @@ manager command


ylinzhu (approved):
ok

PR #2681, opened 2021-05-21
inner-1138:remove null field in json

dcy10000:

Reason:  
  inner-1138:remove null field in json
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #2682, opened 2021-05-21
 cherry pick from pr(2667) and  fix rwSplitMode check inner 1081

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG 
Influences：  
   fix rwSplitMode check inner 1081


ylinzhu:
already cherry pick so close

PR #2683, opened 2021-05-24
Optimize implicit submission of sharding's ddl

wenyh1:
Reason:  
  BUG  inner 917、1110、982
Type:  
  BUG
Influences：  
   Optimize implicit submission of sharding's ddl


LUAgam (approved):
lgtm

PR #2684, opened 2021-05-24
inner-1136: can bootstrap if sharding.xml is empty

dcy10000:

Reason:  
  BUG inner-1136: can bootstrap if sharding.xml is empty
Type:  
  BUG
Influences：  
   bootstrap


LUAgam (approved):
ok

PR #2685, opened 2021-05-24
fix: sql statistics for NPE

wenyh1:
Reason:  
  BUG inner 1141
Type:  
  BUG
Influences：  
   fix：sql statistics for NPE


LUAgam (approved):
ok

PR #2686, opened 2021-05-24
fix load data  incorrect execution result inner 1098

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix load data  incorrect execution result inner 1098


LUAgam (approved):
ok

PR #2687, opened 2021-05-24
fix service ClassCastException using dble rwSplit

PanternBao:
Reason:  
  BUG #inner 1133.
Type:  
  BUG
Influences：  
  fix service ClassCastException using dble rwSplit


dcy10000 (approved):
ok

PR #2688, opened 2021-05-24
fix  use the same method to deal with property inner 920

ylinzhu:
Reason:  
  BUG inner 920
Type:  
  BUG
Influences：  
   fix  use the same method to deal with property inner 920


dcy10000 (approved):
ok

PR #2689, opened 2021-05-25
inner-1139: move ddl trace to another  log file

dcy10000:

Reason:  
  BUG inner-1139: move ddl trace to another  log file
Type:  
  BUG
Influences：  
  trace log


ylinzhu (approved):
ok

PR #2690, opened 2021-05-25
fix some bug about sql digest

PanternBao:
Reason:  
  BUG.  
Type:  
  BUG
Influences：  
  fix some bug about sql digest


LUAgam (approved):
ok

PR #2691, opened 2021-05-26
handle error packet in backend handshake phase to avoid ArrayIndexOutOfBoundException

PanternBao:
Reason:  
  BUG #inner 1125.  
Type:  
  BUG
Influences：  
   handle error packet in backend handshake phase to avoid ArrayIndexOutOfBoundException


wenyh1 (approved):
lgtm

PR #2692, opened 2021-05-26
fix filter conditions like group by, order by ,having inner 911

ylinzhu:
Reason:  
  BUG  inner 911, inner 1150
Type:  
  BUG
Influences：  
   fix filter conditions like group by, order by ,having inner 911


dcy10000 (approved):
lgtm

PR #2697, opened 2021-05-27
fix: connection affinity in a transaction

LUAgam:
Reason:  
  BUG inner-936
Type:  
  BUG
Influences：  
fix: connection affinity in a transaction


PanternBao (approved):
lgtm

PR #2698, opened 2021-05-27
fix: adjust the sequence of locks

LUAgam:
Reason:  
  BUG inner-950
Type:  
  BUG
Influences：  
fix: adjust the sequence of locks


ylinzhu (approved):
ok

PR #2701, opened 2021-05-27
Inner 1128   organize the write  method and implement a general callback

dcy10000:
Reason:  
  Improve Inner 1128    organize the write  method and implement a general callback
Type:  
  Improve  
Influences：  
      write


dcy10000:
delayed because of ci fail

ylinzhu (approved):
ok

PR #2702, opened 2021-05-27
inner-1154: reload config should succeed if cluster meta is empty

dcy10000:

Reason:  
  BUG inner-1154: reload config should succeed if cluster meta is empty
Type:  
  BUG
Influences：  
   cluster waiting


wenyh1 (approved):
lgtm

PR #2705, opened 2021-05-28
fix check tableName inner 1104

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix check tableName inner 1104


LUAgam (approved):
ok

PR #2707, opened 2021-05-28
fix: rule verification should be before acquiring the lock

LUAgam:
Reason:  
  BUG inner-1063
Type:  
  BUG
Influences：  
fix: rule verification should be before acquiring the lock


ylinzhu (approved):
ok

LUAgam:
merge conflict, resubmit

PR #2717, opened 2021-06-01
fix: the DDL examined_rows value is not counted correctly

wenyh1:
Reason:  
  BUG inner 1164
Type:  
  BUG
Influences：  
   fix xx


dcy10000 (approved):
ok

PR #2720, opened 2021-06-02
fix drop view sql hang  and add checkView

ylinzhu:
Reason:  
  BUG inner 1163 ,inner 1171
Type:  
  BUG
Influences：  
 fix drop view sql hang  and add checkView


PanternBao (approved):
lgtm

PR #2724, opened 2021-06-02
fix: execute multi-sql,statistics table is not correct

wenyh1:
Reason:  
  BUG inner 1173
Type:  
  BUG
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #2728, opened 2021-06-04
make start_time in sql_log manager table human readable

PanternBao:
Reason:  
  BUG #inner 994.
Type:  
  BUG
Influences：  
  make start_time in sql_log manager table human readable


wenyh1 (approved):
ok

PR #2729, opened 2021-06-04
fix: when a connection exits directly in a transaction, the statistics table accidentally loses the transaction

wenyh1:
…s table accidentally loses the transaction

Reason:  
  BUG inner 1186
Type:  
  BUG
Influences：  
   fix: when a connection exits directly in a transaction, the statistics table accidentally loses the transaction


PanternBao (approved):
lgtm

PR #2730, opened 2021-06-07
fix: rule verification should be before acquiring the lock(cherry pick from 2725)

LUAgam:
Reason:
  BUG inner-1063
Type:
  BUG
Influences：
fix: rule verification should be before acquiring the lock
#2725


ylinzhu (approved):
ok

PR #2733, opened 2021-06-08
fix: xa-id adds table information

LUAgam:
Reason:  
  BUG inner-1188
Type:  
  BUG
Influences：  
* xa-id adds table information
* user-max_conn_count default value is 0


ylinzhu (approved):
ok

PR #2734, opened 2021-06-09
handle duplicate errorCode

wenyh1:
Reason:  
  BUG inner 1155
Type:  
  BUG
Influences：  
   handle duplicate errorCode


ylinzhu (approved):
ok

PR #2735, opened 2021-06-09
subquery  push down  inner 1192

ylinzhu:
Reason:  
  BUG  inner 1192
Type:  
  BUG
Influences：  
  subquery  push down  inner 1192


LUAgam (approved):
ok

PR #2737, opened 2021-06-10
fix: when using split fields to associate tables, try to optimize

LUAgam:
Reason:  
  BUG inner-1193
Type:  
  BUG
Influences：  
fix: when using split fields to associate tables, try to optimize


wenyh1 (approved):
lgtm

PR #2739, opened 2021-06-11
inner-960-optimize:reduce logs and adjust delay logic.

dcy10000:
Reason:  
   Improve:inner-960-optimize:reduce logs and adjust delay logic.
Type:  
  Improve
Influences：  
   graceful close


wenyh1 (approved):
lgtm

PR #2740, opened 2021-06-11
inner-940:avoid that backend connection is closed before setting ResponseHandler to backend connection

PanternBao:
Reason:  
  BUG inner#940.
Type:  
  BUG
Influences：  
  avoid that backend connection is closed before setting ResponseHandler to backend connection


dcy10000 (approved):
ok

PR #2741, opened 2021-06-11
fix NPE, print stack adjustment

wenyh1:
Reason:  
  BUG inner 1186
Type:  
  BUG
Influences：  
   fix NPE, print stack adjustment


dcy10000 (approved):
ok

dcy10000 (commented):
use @NonNull annotation could be better

PR #2742, opened 2021-06-15
SQL statistics should not affect the business logic

wenyh1:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   SQL statistics should not affect the business logic


dcy10000 (approved):
lgtm

PR #2743, opened 2021-06-16
fix: lock degradation of dble_information

LUAgam:
Reason:  
  BUG inner-1212/inner-1217/inner-1061
Type:  
  BUG
Influences：  
fix: lock degradation of dble_information
fix npe

### DIFF
--- a/src/main/java/com/actiontech/dble/services/manager/handler/DeleteHandler.java
+++ b/src/main/java/com/actiontech/dble/services/manager/handler/DeleteHandler.java
@@ -5,6 +5,7 @@
 
 package com.actiontech.dble.services.manager.handler;
 
+import com.actiontech.dble.DbleServer;
 import com.actiontech.dble.cluster.ClusterHelper;
 import com.actiontech.dble.cluster.DistributeLock;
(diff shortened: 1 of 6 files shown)


src/main/java/com/actiontech/dble/services/manager/handler/DeleteHandler.java
```diff
@@ -106,7 +108,8 @@ public void handle(String stmt, ManagerService service) {
         ManagerWritableTable managerTable = (ManagerWritableTable) managerBaseTable;
         //stand-alone lock
         int rowSize;
-        boolean lockFlag = managerTable.getLock().tryLock();
+        final ReentrantReadWriteLock lock = DbleServer.getInstance().getConfig().getLock();
+        boolean lockFlag = lock.writeLock().tryLock();
         if (!lockFlag) {
             service.writeErrMessage(ErrorCode.ER_YES, "Other threads are executing management commands(insert/update/delete), please try again later.");
```
dcy10000, line 114: need adjust log because the lock is changed

dcy10000 (approved):
lgtm

PR #2748, opened 2021-06-18
fix: unlock should be after return the packet

LUAgam:
Reason:  
  BUG inner-1211/inner-1212
Type:  
  BUG
Influences：  
fix: unlock should be after return the packet


dcy10000 (approved):
lgtm

PR #2752, opened 2021-06-25
fix: adjust the order of fixField and modify dble_thread_pool

LUAgam:
Reason:  
  BUG inner-894/inner-1226
Type:  
  BUG
Influences：  
fix: adjust the order of fixField and modify dble_thread_pool


dcy10000 (approved):
ok

PR #2753, opened 2021-06-25
fix:the closing logic is not as expected(cherry pick inner 1225)

wenyh1:
Reason:  
  BUG 1225
Type:  
  BUG
Influences：  
   fix:the closing logic is not as expected(cherry pick inner 1225)


PanternBao (approved):
lgtm

PR #2755, opened 2021-06-25
fix  sub query too much rows inner 1165

ylinzhu:
Reason:  
  BUG inner 1165
Type:  
  BUG
Influences：  
  fix  sub query too much rows inner 1165


PanternBao (approved):
lgtm

PR #2757, opened 2021-06-25
inner-1234:move beforeHandlingTask to prevent NPE when using multiPacket .

dcy10000:

Reason:  
  BUG inner-1234:move beforeHandlingTask to prevent NPE when using multiPacket .
Type:  
  BUG
Influences：  
   trace +Statistic


ylinzhu (approved):
ok

PR #2758, opened 2021-06-29
inner-1187: fix error when column  contains subquery

dcy10000:

Reason:  
  BUG inner-1187: fix error when a column contains a subquery
Type:  
  BUG
Influences：  
   the scenario that contains subqueries


wenyh1 (approved):
lgtm

PR #2761, opened 2021-07-02
inner-1239:avoid thread leaking for group by

PanternBao:
Reason:  
  BUG #inner 1239.
Type:  
  BUG
Influences：  
  avoid thread remaining for group by


LUAgam (approved):
lgtm

PR #2763, opened 2021-07-02
deal with property inner 920

ylinzhu:
Reason:  
  Improve 
Type:  
  Improve  
Influences：  
  deal with property inner 920


PanternBao (approved):
lgtm

PR #2764, opened 2021-07-05
fix zk starts the process inner 1243

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix zk starts the process inner 1243


PanternBao (approved):
lgtm

PR #2765, opened 2021-07-05
🐛  :fix some NPE when concurrent close happens.

dcy10000:
Reason:  
  BUG bugfix :fix some NPE when concurrent close happens.
Type:  
  BUG/Improve  
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #2766, opened 2021-07-05
inner-1244:fix connection.isValid() cause error

dcy10000:

Reason:  
  BUG inner-1244:fix connection.isValid() cause error
Type:  
  BUG
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #2769, opened 2021-07-06
fix: parameter verification/sync map

LUAgam:
Reason:  
  BUG inner-1245/inner-1226
Type:  
  BUG
Influences：  
fix: parameter verification/sync map


ylinzhu (approved):
ok

PR #2774, opened 2021-07-13
inner-1252: add default method for buffer write

dcy10000:

Reason:  
  BUG inner-1252: add default method for buffer write
Type:  
  BUG
Influences：  
   write EOF packet


wenyh1 (approved):
lgtm

PR #2779, opened 2021-07-15
fix load data incomplete data import of multiple nodes inner 1256

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix load data incomplete data import of multiple nodes inner 1256


dcy10000 (approved):
lgtm

PR #2785, opened 2021-07-19
'txInterrupted' is not updated when an implicit commit statement is executed in error

wenyh1:
Reason:  
  BUG inner 1261
Type:  
  BUG
Influences：  
   1、 'txInterrupted' is not updated when an implicit commit statement is executed in error
         2、Adjust the class(LockTablesHandler/MysqlCreateViewHandler/MysqlDropViewHandler ..)


ylinzhu (approved):
ok

PR #2786, opened 2021-07-19
fix: error message adjustment/sync map

LUAgam:
Reason:  
  BUG inner-1260/inner-1226
Type:  
  BUG
Influences：  
fix: error message adjustment/sync map


PanternBao (approved):
lgtm

PR #2789, opened 2021-07-19
fix bug about group_concat order by

PanternBao:
Reason:  
  BUG #inner 1220.
Type:  
  BUG 
Influences：  
  fix bug about group_concat order by


LUAgam (approved):
ok

PR #2790, opened 2021-07-20
fix: cannot find view with schema

wenyh1:
Reason:  
  BUG inner 1264
Type:  
  BUG 
Influences：  
   fix: cannot find view with schema


PanternBao (approved):
lgtm

PR #2792, opened 2021-07-22
fix: reference startExecuteBackend method

LUAgam:
Reason:  
  BUG inner-1271
Type:  
  BUG
Influences：  
fix: reference startExecuteBackend method


dcy10000 (approved):
lgtm

PR #2795, opened 2021-07-22
fix: reference startExecuteBackend method/NPE

LUAgam:
Reason:  
  BUG inner-1270/inner-1271
Type:  
  BUG
Influences：  
fix: reference startExecuteBackend method/NPE


PanternBao (approved):
lgtm

PR #2797, opened 2021-07-23
fix  zk view initialization failed inner 1243

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG  
Influences：  
  fix  zk view initialization failed inner 1243


LUAgam (approved):
ok

PR #2799, opened 2021-07-23
fix shardingNode have fake dbGroup need to start failure inner 1265

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix shardingNode have fake dbGroup need to start failure inner 1265


wenyh1 (approved):
lgtm

PR #2801, opened 2021-07-26
fix: lost connection occurs when creating (mysql) view

wenyh1:
Reason:  
  BUG inner 1279
Type:  
  BUG
Influences：  
   fix: lost connection occurs when creating (mysql) view


ylinzhu (approved):
ok

PR #2802, opened 2021-07-26
inner-1280: fix miss id in lower_case_table_name clone

dcy10000:
Reason:  
  BUG inner-1280: fix miss id in lower_case_table_name clone
Type:  
  BUG
Influences：  
  id of table config


wenyh1 (approved):
lgtm

PR #2804, opened 2021-07-27
fix zk create view not have view lock inner 1281

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix zk create view not have view lock inner 1281


LUAgam (approved):
OK

PR #2806, opened 2021-07-28
cherry pick from pr(2804) and fix zk create view not have view lock inner 1281 

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG


ylinzhu:
the modifications are simple and verified on master

PR #2809, opened 2021-07-30
fix checkTableConsistency error when dble metadata is the same as mysql metadata inner 1276

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
ok

PR #2811, opened 2021-08-03
fix  npe  when zk fisrt start  inner 1292

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix  npe  when zk fisrt start  inner 1292


ylinzhu:
Ci does not work and no commit record exists cannot be rebuild

PR #2812, opened 2021-08-03
cherry pick from pr(2809) amd fix checkTableConsistency error when dble metadata is the same as mysql metadata

ylinzhu:
Reason:  
  BUG  
Type:  
  BUG
Influences：  
  fix checkTableConsistency error when dble metadata is the same as mysql metadata


ylinzhu:
The master has been validated and the logic is not complex

PR #2815, opened 2021-08-03
fix management execution sql will have conversion errors inner 1293

ylinzhu:
Reason:  
  BUG
Type:  
  BUG
Influences：  
  fix management execution sql will have conversion errors inner 1293


wenyh1 (approved):
lgtm

PR #2818, opened 2021-08-05
fix: requests are evenly distributed to io threads(cherry-pick)

LUAgam:
Reason:  
  BUG inner-1296
Type:  
  BUG
Influences：  
fix: requests are evenly distributed to io threads


ylinzhu (approved):
ok

PR #2820, opened 2021-08-09
inner-1300:replace from copyResource.sh to maven profile

dcy10000:

Reason:  
  Improve inner-1300:replace from copyResource.sh to maven profile
Type:  
  Improve
Influences：  
       develop


wenyh1 (approved):
lgtm

PR #2831, opened 2021-08-24
remove useless dependency in pom.xml

PanternBao:
Reason:  
  Improve.  
Type:  
     Improve  
Influences：  
  remove useless dependency in pom.xml


yanhuqing666 (approved):
ok

PR #2835, opened 2021-09-01
inner 1284 - add log for manager command

PanternBao:
Reason:  
  Improve #inner 1284.  
Type:  
  Improve  
Influences：  
  add log for manager command


LUAgam (approved):
lgtm

PR #2839, opened 2021-09-02
inner-1002:support detach/attach cluster

dcy10000:
Reason:  
   Improve inner-1002:support detach/attach cluster
Type:  
     Improve  
Influences：  
   1. support detach/attach cluster
   2. implement general delay logic


PanternBao (dismissed):
lgtm

PR #2840, opened 2021-09-03
#2611 child table no need finish route

yanhuqing666:
Reason:  
  BUG #2611
Type:  
  BUG 
Influences：  
   fix childtable finish route


LUAgam (approved):
lgtm

PR #2841, opened 2021-09-03
fix: sorting to eliminate thread interference

LUAgam:
Reason:  
  BUG inner-1345
Type:  
  BUG
Influences：  
fix: sorting to eliminate thread interference


ylinzhu (approved):
ok

PR #2852, opened 2021-09-09
inner-1365&1364&1366:1. ha command should detached ,too. 2. resolve pause hang when detached

dcy10000:

Reason:  
  BUG inner-1365&1364:1. ha command should detached ,too. 2. resolve pause hang when detached
Type:  
  BUG
Influences：  
   cluster detach


PanternBao (approved):
lgtm

PR #2853, opened 2021-09-09
inner 1361 - fix that some manager command doesn't support comment

PanternBao:
Reason:  
  BUG #inner 1361.
Type:  
  BUG
Influences：  
  fix that some manager command doesn't support comment


wenyh1 (approved):
lgtm

PR #2856, opened 2021-09-09
fix: connection failed during handshake phase did not count 'totalConnections' correctly

wenyh1:
Reason:  
  BUG inner 1306
Type:  
  BUG
Influences：  
   fix: connection failed during handshake phase did not count 'totalConnections' correctly


LUAgam (approved):
ok

PR #2858, opened 2021-09-10
support jdk 11 in maven

PanternBao:
Reason:  
  Improve #.  
Type:  
  Improve  
Influences：  
  support jdk 11 in maven


ylinzhu (approved):
ok

PR #2859, opened 2021-09-13
fix some unsupported manager command due to inner#1284

PanternBao:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix some unsupported manager command due to inner#1284


LUAgam (approved):
ok

PR #2860, opened 2021-09-13
fix: prompt optimization

LUAgam:
Reason:  
  BUG inner-1266
Type:  
  BUG
Influences：  
fix: prompt optimization


PanternBao (approved):
ok

PR #2861, opened 2021-09-13
support jdk11 - remove system properties of jdk11 from system config

PanternBao:
Reason:  
  Improve.  
Type:  
    Improve  
Influences：  
    support jdk11 - remove system properties of jdk11 from system config


dcy10000 (approved):
lgtm

PR #2863, opened 2021-09-14
support comment for some manager command

PanternBao:
Reason:  
   Improve inner# 1284.  
Type:  
  Improve  
Influences：  
  support comment for some manager command


LUAgam (approved):
ok

PR #2867, opened 2021-09-16
fix heartbeat deadlock inner  1372

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG  
Influences：  
  fix heartbeat deadlock inner  1372


dcy10000 (approved):
ok

PR #2870, opened 2021-09-16
inner-1387: fix closeTimeout doesn't set

dcy10000:

Reason:  
  Bug inner-1387: fix closeTimeout doesn't set
Type:  
  BUG
Influences：  
   fix xx


PanternBao (approved):
ok

PR #2875, opened 2021-09-17
inner 1351 - update reloading timestamp after reload @@metadata

PanternBao:
Reason:  
  BUG #inner 1351.
Type:  
  BUG
Influences：  
  update reloading timestamp after reload @@metadata


LUAgam (approved):
ok

PR #2880, opened 2021-09-17
fix: avoid the soaring cpu caused by nio-select

LUAgam:
Reason:  
  BUG #11/inner-1304
Type:  
  BUG
Influences：  
fix: avoid the soaring cpu caused by nio-select
jdk-bug：https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6403933
refer to https://github.com/netty/netty/blob/3.10/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java


PanternBao (approved):
lgtm

PR #2882, opened 2021-09-18
fix manager sql error inner 1381

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
 fix manager sql error inner 1381


PanternBao (approved):
lgtm

PR #2884, opened 2021-09-18
An alarm is generated when tables are inconsistent during startup

wenyh1:
Reason:  
  BUG inner 1392、1395
Type:  
  BUG
Influences：  
   fix：1、An alarm is generated when tables are inconsistent during startup
   2、Optimization hints


src/main/java/com/actiontech/dble/meta/table/DefaultNodeTableHandler.java
```diff
@@ -121,7 +121,7 @@ else if (!localTables.containsKey(table))
         private final AbstractSchemaMetaHandler operationalHandler;
         private final String schema;
         private final SchemaConfig schemaConfig;
-        private final Map<String, Map<String, List<String>>> tablesStructMap = new HashMap<>();
+        private volatile Map<String, Map<String, List<String>>> tablesStructMap = new HashMap<>();
```
dcy10000, line 124: why?

PR #2889, opened 2021-09-23
inner-1294: fix flow control doesn't work when close suddenly

dcy10000:

Reason:  
  BUG  fix flow control doesn't work when close suddenly
Type:  
  BUG
Influences：  
   flowcontrol


LUAgam (approved):
ok

PR #2891, opened 2021-09-23
fix that errorprone is slow

PanternBao:
Reason:  
  Improve.  
Type:  
  Improve  
Influences：  
  fix that errorprone is slow


wenyh1 (approved):
lgtm

PR #2893, opened 2021-09-23
enhanced create syntax in manager command

PanternBao:
Reason:  
  Improve.  
Type:  
  Improve  
Influences：  
  enhanced create syntax in manager command


wenyh1 (approved):
lgtm

PR #2898, opened 2021-09-26
inner-1308:fix cleanup operations  trigger simultaneously.

dcy10000:

Reason:  
  BUG inner-1308:fix cleanup operations trigger simultaneously.
Type:  
  BUG
Influences：  
    cleanup  when close


LUAgam (approved):
ok

PR #2902, opened 2021-09-27
An alarm is generated when tables are inconsistent

wenyh1:
Reason:
  BUG inner 1392、1395
Type:
  BUG
Influences：
   fix：1、An alarm is generated when tables are inconsistent during startup
   2、Optimization hints


dcy10000 (approved):
ok

PR #2903, opened 2021-09-28
remove resource in pom.xml

PanternBao:
Reason:  
  Improve.  
Type:  
  Improve  
Influences：  
  remove resource in pom.xml


dcy10000 (approved):
ok

PR #2904, opened 2021-09-28
fix  dble need write file in cluster operations inner 1409

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix dble need write file in cluster operations inner 1409


src/main/java/com/actiontech/dble/cluster/ClusterLogic.java
```diff
@@ -856,7 +857,6 @@ public static void syncShardingXmlToLocal(String shardingConfig, XmlProcessBase
 
         //the config Value in ucore is an all in one json config of the sharding.xml
         Shardings sharding = ClusterLogic.parseShardingJsonToBean(gson, shardingConfig);
-        ClusterLogic.writeMapFileAddFunction(sharding.getFunction());
```
yanhuqing666, line 859: why romoved?

ylinzhu: the file has already been written , this is not required

LUAgam (approved):
ok

PR #2905, opened 2021-09-28
fix  dble need write file in cluster operations inner 1409

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
  fix  dble need write file in cluster operations inner 1409


ylinzhu:
More details need to be worked out

LUAgam (approved):
ok

PR #2927, opened 2021-10-22
fix[inner-1429]: strengthen the control onCreateFail

wenyh1:
Reason:  
  BUG inner 1429
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (commented):
wrong comment, ignore

PR #2936, opened 2021-10-29
inner 1430 - fix deadlock between close and backend cleaner

PanternBao:
Reason:  
  BUG #inner 1430.
Type:  
  BUG
Influences：  
  inner 1430 - fix deadlock between close and backend cleaner


ylinzhu (approved):
ok

PR #2937, opened 2021-10-29
inner-1443:fix :memory leak when suddenly close

dcy10000:

Reason:  
  BUG inner-1443::fix :memory leak when suddenly close
Type:  
  BUG
Influences：  
   write logic


PanternBao (approved):
lgtm

PR #2938, opened 2021-11-01
fix READ_LOAD and WRITE_LOAD are always displayed as 0 inner 1445

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
ok

PR #2943, opened 2021-11-01
inner-1446: pause doesn't work in ucore

dcy10000:

Reason:  
  BUG inner-1446: pause doesn't work in ucore
Type:  
  BUG  
Influences：  
  pause in ucore


PanternBao (approved):
lgtm

PR #2947, opened 2021-11-09
fix[inner-1452]: sharding.xml changes are not synchronized to the cluster when switching to pure rwSplit

wenyh1:
Reason:  
  BUG inner 1452
Type:  
  BUG/Improve  
Influences：  
   fix：sharding.xml changes are not synchronized to the cluster when switching to pure rwSplit


wenyh1:
todo discuss

wenyh1:
after discussion, not a bug

PR #2948, opened 2021-11-09
inner 1454 - release serializable lock when front connection is closed

PanternBao:
Reason:  
  BUG #inner 1454.
Type:  
  BUG 
Influences：  
  inner 1454 - release serializable lock when front connection is closed


LUAgam (approved):
ok

PR #2949, opened 2021-11-09
fix[inner-1455]: 'ConcurrentModificationException' occurred while loading metadata

wenyh1:
Reason:  
  BUG inner 1455
Type:  
  BUG
Influences：  
   fix：'ConcurrentModificationException' occurred while loading metadata


PanternBao (approved):
ok

PR #2952, opened 2021-11-10
fix distributed locks not release inner 1456

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #2957, opened 2021-11-16
backend connection in use is returned to connection pool

PanternBao:
Reason:  
  BUG inner1454.  
Type:  
  BUG
Influences：  
  backend connection in use is returned to connection pool


LUAgam (approved):
ok

PR #2959, opened 2021-11-17
fix deadlock between close and manager sql inner 1464

ylinzhu:
Reason:  
  BUG
Type:  
  BUG 
Influences：  
   fix xx


PanternBao:
the problem has been finished in [dead lock pr](https://github.com/actiontech/dble/pull/2936/files)

ylinzhu:
the problem has been resolved in the dead lock pr

PR #2960, opened 2021-11-17
improve: optimize split

LUAgam:
Reason:  
  Improve inner-1444
Type:  
  Improve  
Influences：  
fix: optimize split


ylinzhu (approved):
ok

PR #2962, opened 2021-11-19
cherry pick :fix memory leaks on inner-1461/inner-1462/inner-1457

dcy10000:
Reason:  
  BUG cherry pick :fix memory leaks on inner-1461/inner-1462/inner-1457
Type:  
  BUG  
Influences：  
   direct memory


wenyh1 (approved):
lgtm

PR #2965, opened 2021-11-24
Standardize 'group by' pushdown to MySQL and fix #1451 TableNode group by push down do not change to alias

### LINKED ISSUE #1451: sql "select a as b from table group by a " returns error when b is one of the table's column names
* **dble version：version: 5.6.23-dble-2.19.07.0-dd160d6138a16228617b75a20d284dd117f5b138-20190919114227**  
* **preconditions ：**  
```
create table test (area_id int ,areaId int ,name char(10));
insert into test values(1,2,'q'),(4,5,'f'),(7,8,'h');

```
* **configs：** 

**schema.xml**  

```
 <schema name="testdb" dataNode="dn1">
        <table name="test" rule="hash-4" dataNode="dn1,dn2,dn3,dn4"></table>
    </schema>

```

**rule.xml**  

```
   <tableRule name="hash-4">
        <rule>
            <columns>area_id</columns>
            <algorithm>4-long</algorithm>
        </rule>
    </tableRule>
     <function name="4-long" class="Hash">
        <property name="partitionCount">4</property>
        <property name="partitionLength">1</property>
    </function>
```

* **steps：**    
1.execute in dble server port with alias is one of the table's column names:
```
mysql>  select area_id as areaId,count(1) as count from test group by area_id;
ERROR 1003 (HY000): Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'db4.test.area_id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```
2.change the alias to other
```
mysql>  select area_id as areaIda,count(1) as count from test group by area_id;
+---------+-------+
| areaIda | count |
+---------+-------+
|       1 |     1 |
|       4 |     1 |
|       7 |     1 |
+---------+-------+
3 rows in set (0.00 sec)
```
3.exeute the same sql as step1's  in mysql and return correct result
```
mysql>  select area_id as areaId,count(1) as count from test group by area_id;
+--------+-------+
| areaId | count |
+--------+-------+
|      4 |     1 |
+--------+-------+
1 row in set (0.00 sec)
```

* **expect result：**  
    1.step 1 should retuen correct result
* **real result：**  
    1.step 1 return error

LavieZhou:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx

### DIFF
--- a/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/builder/sqlvisitor/PushDownVisitor.java
+++ b/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/builder/sqlvisitor/PushDownVisitor.java
@@ -293,26 +293,23 @@ public class PushDownVisitor extends MysqlVisitor {
                             } else {
                                 pdName = visitUnSelPushDownName(groupCol, true);
                             }
-                        sqlBuilder.append(pdName).append(" ").append(group.getSortOrder()).append(",");
+                        sqlBuilder.append(pdName).append(",");
                     }
                     sqlBuilder.deleteCharAt(sqlBuilder.length() - 1);
                 }
-            } else {
-                // change to push down order by
-                pushDownOrderBy.addAll(query.getGroupBys());
-                if (pushDownOrderBy.size() > 0) {
-                    sqlBuilder.append(" ORDER BY ");
-                    for (Order order : pushDownOrderBy) {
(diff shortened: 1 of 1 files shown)


PanternBao (approved):
LGTM

PR #2972, opened 2021-11-29
fix bug that dble can't find null constant column in select list

PanternBao:
Reason:  
  BUG  
Type:  
  BUG
Influences：  
  fix : select null as aaa from sharding_a union select null as bbb from sharding_b


yanhuqing666 (approved):
lgtm

PR #2973, opened 2021-11-29
fix ClassCastException when splitting or condition

PanternBao:
Reason:  
  BUG.
Type:  
  BUG
Influences：  
  fix ClassCastException when splitting or condition


yanhuqing666 (approved):
ok

PR #2976, opened 2021-12-01
inner-1471: modify join order use hint

dcy10000:

Reason:  
  BUG inner-1471: modify join order use hint
Type:  
  BUG
Influences：  
   join order


yanhuqing666 (approved):
changes LGTM
others review  later

PR #2977, opened 2021-12-01
abort optimize when Cartesian with relation

yanhuqing666:
Reason:  
  inner 1494
Type:  
  BUG 
Influences：  
   abort optimize when Cartesian with relation


dcy10000 (approved):
lgtm

PR #2979, opened 2021-12-02
fix: exception handling & dos newline(\r) handling

LUAgam:
Reason:  
  BUG inner-1496
Type:  
  BUG
Influences：  
fix: exception handling
queue default value modification
dos newline handling 
for druid does not support the processing of \r


PanternBao (approved):
lgtm

PR #2980, opened 2021-12-03
inner-1415: pause return two error message

dcy10000:

Reason:  
  BUG inner-1415: pause return two error message
Type:  
  BUG

Influences：  
   pause


wenyh1 (approved):
lgtm

Issue #2981, opened 2021-12-04
The field value has a semicolon, SQL sometimes parses incorrectly and sometimes parses normally


LavieZhou:
* **dble version：2.19.03/lts**  
* **preconditions ：**  
no
* **configs：** 

**schema.xml**  

```


```

**rule.xml**  

```


```

**server.xml**  

```


```

* **steps：**    
    step1.  execute the following code
[TestDurid.txt](https://github.com/actiontech/dble/files/7653473/TestDurid.txt)
* **expect result：**  
    1.sql1 is UPDATE IPVS_REMITSNAP 
SET 
REPAMT2 = '0',
REPPS2 = NULL,
ISREF = 'N',
CONTRNO = 'PO-2021-08-0026',
INVOINO = 'CSJ-184/21;',
REGNO = NULL
WHERE REPCODE = '110101000248211123CG0A

sql2 is UPDATE IPVS_REMITSNAP SET REPAMT2 = '0',REPPS2 = NULL,ISREF = 'N',CONTRNO = 'PO-2021-08-0026',INVOINO = 'CSJ-184/21;',REGNO = NULL WHERE REPCODE = '110101000248211123CG0A

* **real result：**  
    1.sql1 is UPDATE IPVS_REMITSNAP 
SET 
REPAMT2 = '0',
REPPS2 = NULL,
ISREF = 'N',
CONTRNO = 'PO-2021-08-0026',
INVOINO = 'CSJ-184/21

sql2 is UPDATE IPVS_REMITSNAP SET REPAMT2 = '0',REPPS2 = NULL,ISREF = 'N',CONTRNO = 'PO-2021-08-0026',INVOINO = 'CSJ-184/21;',REGNO = NULL WHERE REPCODE = '110101000248211123CG0A

* **supplements：**  
    1.
![image](https://user-images.githubusercontent.com/46223149/144700358-7ec764b5-c14c-40df-a6ce-a91f6f181c3b.png)
The code should return the initial value of offset, it is recommended to modify 
![image](https://user-images.githubusercontent.com/46223149/144700392-931dede4-6ccf-4c41-97aa-1a0c0dc5673f.png)

PanternBao:
The lastest version(eg, 3.21.10.0) has fix the issue

PR #2984, opened 2021-12-08
issue 2981 - parse multi sql error

PanternBao:
Reason:  
  BUG #2981.
Type:  
  BUG
Influences：  
  parse multi sql error


yanhuqing666 (approved):
ok

PR #2990, opened 2021-12-10
fix: the split command will not be controlled by idleTime

LUAgam:
Reason:  
  BUG inner-1510
Type:  
  BUG
Influences：  
fix: the split command will not be controlled by idleTime


PanternBao (approved):
lgtm

PR #2991, opened 2021-12-10
Select item optimizer

yanhuqing666:
Reason:  
   Improve inner 1508
Type:  
   Improve  
Influences：  
   1. fix which import from [3 table union order](https://github.com/actiontech/dble/commit/5a1275fd79b69e4033b92e318e1b8763069be97f#diff-ca61a3fd07652b1b6d9d7f49d49c01bbac574cea17e4160737db5d7bbb160eb2)
   2. if  table-node's function is not an aggregate function,  it will be pushed down


yanhuqing666:
if use old version, 
need to add a line after `Item func = sel.cloneStruct();` ,line 167 of PlanUtil.java
            `func.setAlias(sel.getAlias());`

dcy10000 (approved):
lgtm

PR #2995, opened 2021-12-13
inner 1512 keep order

yanhuqing666:
Reason:  
  BUG inner 1512 
Type:  
  BUG 
Influences：  
   join order will not change


dcy10000 (approved):
ok

PR #2999, opened 2021-12-13
fix: the split command will not be controlled by idleTime

LUAgam:
Reason:
  BUG inner-1510
Type:
  BUG
Influences：
fix: the split command will not be controlled by idleTime
fix: connection skipIdleCheck status reset


wenyh1 (approved):
lgtm

PR #3007, opened 2021-12-15
support non left-deep join trees, eg:right->left

yanhuqing666:
Reason:  
  BUG inner 1528 
Type:  
  BUG 
Influences：  
   support non left-deep join trees, eg:right->left


dcy10000 (approved):
lgtm

PR #3016, opened 2021-12-20
inner 1541 - execute sql that contains comment hang

PanternBao:
Reason:  
  BUG inner#1541.
Type:  
  BUG  
Influences：  
  execute sql that contains comment hang,eg: /* xxx*/ select * from test


LUAgam (approved):
ok

PR #3017, opened 2021-12-20
fix: the ddl execution results are pushed twice in the cluster

wenyh1:
Reason:  
  BUG inner 1543
Type:  
  BUG
Influences：  
   fix: the ddl execution results are pushed twice in the cluster


yanhuqing666 (approved):
lgtm

PR #3020, opened 2021-12-21
add close conf for record heartbeat

yanhuqing666:
Reason:  
   Improve to avoid big records 
Type:  
   Improve  
Influences：  
   manager cmd


dcy10000 (approved):
ok

PR #3022, opened 2021-12-21
fix:column name parsing error

LUAgam:
Reason:  
  BUG inner-1549
Type:  
  BUG
Influences：  
fix:column name parsing error


wenyh1 (approved):
ok

PR #3027, opened 2021-12-23
fix:column name parsing error(cherry pick from 3.21.02.99)

LUAgam:
Reason:
  BUG inner-1549
Type:
  BUG
Influences：
fix:column name parsing error
#3022


wenyh1 (approved):
ok

PR #3028, opened 2021-12-23
cherrypick:inner-1544: fix LVS/SLB heartbeat cause memory leak & inner-1548: fix rounding problem in show @@directmemory 

dcy10000:
Reason:  
  BUG inner-1544: fix LVS/SLB heartbeat cause memory leak 
inner-1548: fix rounding problem in show @@directmemory 
Type:  
  BUG
Influences：  
   directmemory


PanternBao (approved):
lgtm

PR #3030, opened 2021-12-24
1508 chinese col name

yanhuqing666:
Reason:  
  BUG Chinese column  name/alias
Type:  
  BUG
Influences：  
   fix xx


dcy10000 (approved):
ok

PR #3034, opened 2021-12-28
fix failed to release the distributed lock inner 1515

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
lgtm

PR #3037, opened 2021-12-29
field charset

yanhuqing666:
Reason:  
  BUG inner 1560/1565
Type:  
  BUG 
Influences：  
   fix field charset


yanhuqing666:
Because of the large impact, we waited for other PRs to stabilize before merging, temporarily serving as a draft

PR #3041, opened 2021-12-30
[inner-1473]Hint specifies left/right inner replaced by inner join

wenyh1:
Reason:  
  Improve inner 1473 
Type:  
  Improve  
Influences：  
   fix xx


dcy10000 (changes requested):

dcy10000 (approved):
ok

PR #3042, opened 2021-12-30
fix: reselect when dbGroup fails

LUAgam:
Reason:  
  BUG inner-1572
Type:  
  BUG
Influences：  
fix: reselect when dbGroup fails


LUAgam:
need update

PR #3044, opened 2021-12-31
support com_field_list protocol for rw

PanternBao:
Reason:  
  Improve.  
Type:  
  Improve  
Influences：  
  support com_field_list protocol for rw


dcy10000 (approved):
lgtm

PR #3047, opened 2022-01-04
inner-1580:support com_field_list protocol for rw

PanternBao:
Reason:  
   Improve #inner 1580.  
Type:  
  Improve  
Influences：  
  support com_field_list protocol for rw


yanhuqing666 (approved):
lgtm

PR #3050, opened 2022-01-05
[inner-1584] Unreleased lock occurred during the ddl'test-phase when sharding table exception occurred

wenyh1:
Reason:  
  BUG inner 1584
Type:  
  BUG
Influences：  
   fix: Unreleased lock occurred during the ddl'test-phase when sharding table exception occurred


PanternBao (approved):
lgtm

PR #3052, opened 2022-01-05
[inner-1478] create vertical view without falling disk

wenyh1:
Reason:  
  BUG inner-1478
Type:  
  BUG
Influences：  
   fix xx


dcy10000 (approved):
lgtm

PR #3054, opened 2022-01-06
fix:only cancel if the connection is not pending

LUAgam:
Reason:  
  BUG inner-1579
Type:  
  BUG
Influences：  
fix:only cancel if the connection is not pending


wenyh1 (approved):
lgtm

PR #3072, opened 2022-01-12
fix: adjust rw checkSchema to select instance mode

LUAgam:
Reason:  
  BUG innert-1594
Type:  
  BUG
Influences：  
fix: adjust rw checkSchema to select instance mode


ylinzhu (approved):
ok

PR #3073, opened 2022-01-12
inner-1552:add close conf for record heartbeat (#3020)

dcy10000:
Reason:  
  BUG inner-1552:add close conf for record heartbeat 
Type:  
  BUG
Influences：  
   heartbeat


LUAgam (approved):
ok

PR #3074, opened 2022-01-12
fix: csv quote set default

LUAgam:
Reason:  
  BUG inner-1595
Type:  
  BUG
Influences：  
fix: csv quote set default


LUAgam:
need update

PR #3075, opened 2022-01-13
[inner-1543/1498] fix：second push and ddl's hang

wenyh1:
Reason:  
  BUG inner-1543/1498
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #3079, opened 2022-01-13
[inner-1602] fix: to inner join's bug

wenyh1:
Reason:  
  BUG inner 1602
Type:  
  BUG/Improve  
Influences：  
   fix xx


wenyh1:
The repair plan didn’t go through

PR #3081, opened 2022-01-14
[inner-1604] fix: not need to bring 'set autocommit = 0' before executing DDL

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3085, opened 2022-01-17
fix: judging string is empty

LUAgam:
Reason:
  BUG inner-1595
Type:
  BUG
Influences：
fix: judging string is empty


PanternBao (approved):
lgtm

PR #3087, opened 2022-01-17
inner-1427:fix table with alias will route all.

dcy10000:

Reason:  
  BUG inner-1427:fix table with alias will route all.
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #3089, opened 2022-01-17
[inner-1602] fix: where condition missing in single node sql

wenyh1:
Reason:  
  BUG inner 1602
Type:  
  BUG/Improve  
Influences：  
   fix : where condition missing in single node sql


wenyh1:
not thoughtful

wenyh1:
The repair plan didn’t go through

PR #3092, opened 2022-01-18
fix read_load and write_load  displayed error inner 1445

ylinzhu:
Reason:  
  BUG
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #3093, opened 2022-01-18
inner 1475 - strength prepare statement protocol

PanternBao:
Reason:  
  BUG inner 1475. 
Type:  
  BUG
Influences：  
  strength prepare statement protocol


wenyh1 (approved):
lgtm

PR #3094, opened 2022-01-19
[inner-1602] fix: where condition missing in single node sql

wenyh1:
Reason:  
  BUG inner 1602
Type:  
  BUG 
Influences：  
   fix xx


yanhuqing666 (approved):
lgtm

PR #3096, opened 2022-01-20
inner-1613: fix select may send without limit

dcy10000:

Reason:  
  BUG inner-1613: fix select may send without limit
Type:  
  BUG
Influences：  
   select with singleTable


dcy10000:
reopen because of ci error

LUAgam (approved):
ok

PR #3098, opened 2022-01-20
fix: hint does not use sticky connections

LUAgam:
Reason:
  BUG inner-1484
Type:
  BUG
Influences：
fix: hint does not use sticky connections


dcy10000 (approved):
lgtm

PR #3100, opened 2022-01-20
[inner-1612] add release_table_lock stage

wenyh1:
Reason:  
  BUG inner 1612 
Type:  
  BUG 
Influences：  
   fix xx


dcy10000 (approved):
lgtm

Issue #1514, opened 2019-11-26
Every first time sysbench test got [Backend connect Error]


irene-coming:
* **dble version：**  
5.6.29-dble-2.19.09.0-28c8819f535c0fa0405d243628c58ccbff6fde3d-20191120070105
* **preconditions ：**  
from dble -performance-base-testing-of-broadcast-select
* **configs：** 

**schema.xml**  

```


```

**rule.xml**  

```


```

**server.xml**  

```


```

* **steps：**    
    step1.  sysbench test dble:
```sysbench
sysbench --mysql-db=mytest --mysql-host=xx.xx.xx.xx --mysql-port=8066 --mysql-user=xx --mysql-password=xx /usr/share/sysbench/tests/include/oltp_legacy/oltp.lua --oltp_tables_count=1 --oltp-table-size=10000000 --oltp-read-only=on --report-interval=3 --default-charset=utf8 --max-requests=0 --percentile=95 --oltp_skip_trx=on --oltp_point_selects=0 --oltp_simple_ranges=1 --oltp_sum_ranges=0 --oltp_order_ranges=0 --oltp_distinct_ranges=0 --oltp_range_size=9 --threads=512 --time=60 run
```

* **expect result：**  
    1.sysbench success
* **real result：**  
    1.the first time failed for:
```
FATAL: mysql_store_result() returned error 5314 (Backend connect Error, Connection{DataHost[10.186.17.101:3306],Schema[mfdb1]} refused)
FATAL: `thread_run' function failed: /usr/share/sysbench/tests/include/oltp_legacy/oltp.lua:56: db_query() failed
```
 the second time can success.
* **supplements：**  
    1. related dble log:
```log
[...]

Issue #1491, opened 2019-11-12
small int type with PreparedStatement protocol will cause ArrayIndexOutOfBoundsException


yanhuqing666:
* **dble version：all version till to 9.9.9.9 1dc8981d64a9d3930088109f57d5cbd1336863de **  
* **preconditions ：**  
```
drop table if exists test;
create table  test(
id int,
code smallint(3)
) engine =innodb;
insert into test values(1,1);
```

* **configs：** 

**schema.xml**  

```
<table name="test" dataNode="dn1,dn2,dn3,dn4" rule="rule_mod"/>

```

**rule.xml**  

```


```

**server.xml**  

```


```

* **steps：**    
    step1.  
```
public void test4() throws Exception{
		Class.forName("com.mysql.jdbc.Driver");
		Connection connection =  DriverManager.getConnection("jdbc:mysql://127.0.0.1:8066?rewriteBatchedStatements=false&useServerPrepStmts=true", "root", "123456");
		connection.setCatalog("testdb");
		try {
			PreparedStatement stmt = connection.prepareStatement("select id,code from test where id = ?");
			stmt.setInt(1, 1); 
			ResultSet rs = stmt.executeQuery();
			while(rs.next())
			{
				System.out.println(rs.getString(1));
			}
			stmt.clearParameters();
			stmt.close();
		} catch (SQLException e2) {
			System.out.println(e2);
		} finally {
			connection.close();
		}
		
	}
```

* **expect result：**  
    1. code work and print 1
* **real result：**  
    1. java.sql.SQLException: Backend connect Error, Connection{DataHost[xx.xx.xx.xx:xxxx],Schema[xx]} refused
    2. dble.log 
```
[...]

PR #3130, opened 2022-02-14
fix treemap thread safe #1514 (#1518)  and small int #1491

LavieZhou:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3135, opened 2022-02-15
[inner-1615] During ddl execution, the creation and release of a table lock must be the same session on the business side

wenyh1:
Reason:  
  BUG inner 1615
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
ok

PR #3136, opened 2022-02-16
[inner-1615,950] During ddl execution, the creation and release of a table lock must be the same session on the business side (cherry pick)

wenyh1:
Reason:  
  BUG inner 1615,950; similar pr #3135 #2698 
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3137, opened 2022-02-16
[inner-950] adjust the sequence of locks (cherry pick)

wenyh1:
Reason:  
  BUG inner 950；similar pr #2698 
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3139, opened 2022-02-16
[inner-1610] fix：when the subquery is right Join, the result is incorrect

wenyh1:
Reason:  
  BUG inner 1610
Type:  
  BUG/Improve  
Influences：  
   fix when the subquery is right Join, the result is incorrect


dcy10000 (approved):
ok

PR #3146, opened 2022-02-17
[inner-1615] During ddl execution, the creation and release of a table lock must be the same session on the business side (cherry pick)

wenyh1:
Reason:  
  BUG 1615
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3148, opened 2022-02-17
fix where condition miss when try nest loop  inner 1644

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3150, opened 2022-02-23
fix: add split_loaddata

LUAgam:
Reason:  
  Improve inner-1426
Type:  
  Improve  
Influences：  
fix: add split_loaddata


LUAgam:
need update

PR #3155, opened 2022-02-23
inner-1648: report The ER relation limit

dcy10000:

Reason:  
   Improve inner-1648: report The ER relation limit
Type:  
  Improve  
Influences：  
   hint


PanternBao (approved):
ok

PR #3159, opened 2022-02-24
inner-1648-supplement: syntax error when report  The ER relation limit

dcy10000:

Reason:  
   Improve inner-1648-supplement: syntax error when report  The ER relation limit
Type:  
  Improve  
Influences：  
   report syntax


LUAgam (approved):
ok

PR #3161, opened 2022-02-28
fix: support BIN/SUBSTRING function

LUAgam:
Reason:  
  BUG inner-1650
Type:  
  BUG
Influences：  
fix: support BIN/SUBSTRING function
BIN:https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_bin
SUBSTRING:https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_substring


wenyh1 (approved):
ok

PR #3162, opened 2022-03-02
inner-1669: no need to optimize join if the hint order same as execution plan tree

dcy10000:

Reason:  
  BUG inner-1669: no need to optimize join if the hint order same as the execution plan tree
Type:  
  BUG
Influences：  
  hint order


LUAgam (approved):
ok

PR #3163, opened 2022-03-03
inner-1672: column around with `` can't be route

dcy10000:

Reason:  
  BUG inner-1672: column around with `` can't be route
Type:  
  BUG
Influences：  
   route


ylinzhu (approved):
ok

PR #3165, opened 2022-03-03
inner-1666: discard left join when both left join and inner join exists

dcy10000:

Reason:  
  BUG inner-1666: discard left join when both left join and inner join exists
Type:  
  BUG
Influences：  
   join with hint


PanternBao (approved):
lgtm

PR #3169, opened 2022-03-04
[inner-1661] fix: when multiple ER filtering conditions in the join on, query result error

wenyh1:
Reason:  
  BUG inner 1661
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #3170, opened 2022-03-04
Revert "inner-1669: no need to optimize join if the hint order same as execution plan tree (#3162)"

dcy10000:

Reason:  
  BUG Revert "inner-1669: no need to optimize join if the hint order same as execution plan tree (#3162)"
Type:  
  BUG
Influences：  
   join order


ylinzhu (approved):
ok

PR #3172, opened 2022-03-07
[inner-1674] fix: nestloop scenarios's bug

wenyh1:
Reason:  
  BUG inner 1674
Type:  
  BUG
Influences：  
   fix: nestloop scenarios's bug


ylinzhu (approved):
ok

PR #3179, opened 2022-03-11
[inner-1586] during the period of the XA, front and back side accidentally disconnect does not affect the XA retry mechanism

wenyh1:
Reason:  
  BUG inner 1586
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #3180, opened 2022-03-11
[inner-1681] when rwSplitMode="0", don't participate in stickiness

wenyh1:
Reason:  
  BUG inner 1681
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3181, opened 2022-03-11
#3175 - short value of row data protocol fails to convert to  binary protocol

PanternBao:
Reason:  
  BUG #3175. 
Type:  
  BUG
Influences：  
  short value of row data protocol fails to convert to  binary protocol


yanhuqing666 (dismissed):
lgtm

PR #3187, opened 2022-03-16
inner-1667: report an error when using hint with "cartesian with relation"

dcy10000:

Reason:  
  BUG inner-1667: report an error when using hint with "cartesian with relation"
Type:  
  BUG
Influences：  
   cartesian with relation


PanternBao (approved):
lgtm

PR #3194, opened 2022-03-23
fix rwsplit user sqlExecuteTimeout do not take effects  inner 1696

ylinzhu:
Reason:  
  BUG
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #3195, opened 2022-03-24
fix clickhouse Incompatible error inner 1697

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


ylinzhu:
Ci is broken, need reopen

PR #3196, opened 2022-03-28
 fix clickhouse Incompatible error inner 1697 

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG 
Influences：  
   fix xx


dcy10000 (approved):
ok

dcy10000 (changes requested):
has conflict

dcy10000 (approved):

PR #3199, opened 2022-03-29
fix: add backtracking checks

LUAgam:
Reason:  
  BUG inner-1706
Type:  
  BUG
Influences：  
fix: add backtracking checks


wenyh1 (approved):
ok

PR #3200, opened 2022-03-29
avoid deadlock between backend connection close and backend recycle

PanternBao:
Reason:  
  BUG.
Type:  
  BUG
Influences：  
  avoid deadlock between backend connection close and backend recycle


dcy10000 (approved):
lgtm

PR #3207, opened 2022-04-01
[inner-1712&1713] fix：can't find column

wenyh1:
Reason:  
  BUG inner 1712、1713
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #3208, opened 2022-04-01
fix heartbeat timeout and dryrun bug

PanternBao:
Reason:  
  BUG
Type:  
  BUG 
Influences:
        cherry pick from 3.21.10.99


LUAgam (approved):
ok

PR #3209, opened 2022-04-02
[inner-1678] fix: nestloop scenarios's bug

wenyh1:
Reason:  
  BUG inner 1678, similar: #3172 
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3211, opened 2022-04-06
inner 1718 - no need to push down in func

PanternBao:
Reason:  
  BUG inner 1718.
Type:  
  BUG
Influences：  
  no need to push down in func


yanhuqing666 (dismissed):
lgtm

yanhuqing666 (approved):
ok

PR #3212, opened 2022-04-07
inner 1706 - remove schema incorrectly

PanternBao:
Reason:  
  BUG #inner 1706.
Type:  
  BUG
Influences：  
  remove schema incorrectly,eg: table name contains schema name


wenyh1 (approved):
ok

PR #3215, opened 2022-04-12
inner 1721 - fix missing filed attribute

PanternBao:
Reason:  
  BUG #inner 1721.
Type:  
  BUG
Influences：  
   fix missing filed attribute


ylinzhu (approved):
ok

PR #3221, opened 2022-04-15
inner-1666:report error instead of merge left join relations

dcy10000:
Reason:  
  BUG inner-1666:report error instead of merge left join relations
Type:  
  BUG
Influences：  
   join with hint


LUAgam (approved):
ok

PR #3235, opened 2022-04-27
[inner-1661] fix: when multiple ER filtering conditions in the join on, query result error（cherry pick）

wenyh1:
Reason:  
  BUG similar #3169. #3207 
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3239, opened 2022-05-06
[inner-1106] Support for viewing frontend connection activity

wenyh1:
Reason:  
  Improve inner 1106
Type:  
  Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3244, opened 2022-05-10
fix: "between and" returns the wrong table name

LUAgam:
Reason:  
  BUG inner-1746
Type:  
  BUG
Influences：  
fix: "between and" returns the wrong table name


dcy10000 (approved):
lgtm

PR #3250, opened 2022-05-16
fix: xa-id removes shema+table

LUAgam:
Reason:  
  BUG inner-1744
Type:  
  BUG
Influences：  
fix: xa-id removes shema+table


wenyh1 (approved):
lgtm

PR #3260, opened 2022-05-27
fix: adjust the connection acquisition mechanism of the queued threads in the connection pool

LUAgam:
Reason:  
  BUG inner-1751
Type:  
  BUG
Influences：  
fix: adjust the connection acquisition mechanism of the queued threads in the connection pool


ylinzhu (approved):
ok

PR #3262, opened 2022-05-30
fix different types of databases have different check policies inner  1767

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3271, opened 2022-06-08
fix sql error caused by encoding error inner 1411

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


PanternBao (approved):
lgtm

PanternBao (approved):
lgtm

PR #3272, opened 2022-06-09
fix specify a queued thread when creating a new connection

LUAgam:
Reason:  
  BUG inner-1751
Type:  
  BUG
Influences：  
fix specify a queued thread when creating a new connection


ylinzhu (approved):
ok

PR #3275, opened 2022-06-10
fix heartbeat timeout inner 1596 (#3256)

ylinzhu:
Reason:  
  BUG similar: #3256 
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3280, opened 2022-06-13
fix  Insert and delete statements error inner 1787

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


dcy10000 (approved):
ok

PR #3281, opened 2022-06-13
fix  com_field_list unbind the connection inner 1776

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3289, opened 2022-06-16
[inner-1555] cluster 'online' renew thread is not allowed to be killed

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (approved):
ok

PR #3295, opened 2022-06-29
fix when joinStrategyType is 2, an sql execution error occurs inner 1736

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG 
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #3297, opened 2022-06-30
feat: specify a queued thread when creating a new connection（cherry-pick）

LUAgam:
Reason:
  BUG inner-1751
Type:
  BUG
Influences：
fix specify a queued thread when creating a new connection


ylinzhu (approved):
ok

PR #3300, opened 2022-06-30
fix com change user  protocol  error  inner 1807

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG/
Influences：  
   fix nullpoint exception


PanternBao (approved):
ok

PR #3306, opened 2022-07-05
fix analysisUser login with the database name inner 1703

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #3307, opened 2022-07-07
fix general log npe&fix read buffer NPE

dcy10000:

Reason:  
  BUG inner-1789
Type:  
  BUG
Influences：  
  fix general log npe&fix read buffer NPE


LUAgam (approved):
ok

PR #3308, opened 2022-07-07
fix: error getting shard type

LUAgam:
Reason:  
  BUG inner-1820
Type:  
  BUG
Influences：  
fix: error getting shard type


ylinzhu (approved):
ok

PR #3313, opened 2022-07-12
fix automated ci encode format conversion inner 1411

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #3317, opened 2022-07-14
fix: set lowerCase at startup

LUAgam:
Reason:  
  BUG inner-1834
Type:  
  BUG
Influences：  
fix: set lowerCase at startup


dcy10000 (approved):

dcy10000 (approved):
lgtm

PR #3320, opened 2022-07-26
fix: lowerCase initialization should be empty

LUAgam:
Reason:  
  BUG inner-1858
Type:  
  BUG 
Influences：  
fix: lowerCase initialization should be empty


dcy10000 (approved):
lgtm

PR #3321, opened 2022-07-26
[inner-1741] fix：filter conditions for false(boolean) problem

wenyh1:
Reason:  
  BUG 1741
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3330, opened 2022-08-02
[inner-1844] Fix the round/truncate function problem

wenyh1:
Reason:  
  BUG inner 1844
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3331, opened 2022-08-02
inner 1501 - can use table index in complex query hint

PanternBao:
Reason:  
  Improve # inner 1501.  
Type:  
  Improve  
Influences：  
  can use table index in complex query hint


wenyh1 (approved):
lgtm

Issue #3298, opened 2022-06-30
[Improvement] Refine the scope of effect of the reload configuration


LUAgam:
Refine the scope of effect of the reload configuration

Dinosauria:
Verfied in 5.7.21-dble-9.9.9.9-eb2075516306ecdaa656a7da5cd232607d03aa31-20220812020255

LUAgam:
associated documents:
https://actiontech.github.io/dble-docs-cn/history/develop/2.Function/2.001_manager_dble_information.html 
https://actiontech.github.io/dble-docs-cn/history/develop/2.Function/2.19_reload_diff.html

PR #3332, opened 2022-08-02
fix: remove redundant detection after reload

LUAgam:
Reason:  
  BUG inner-1845/inner-1868
Type:  
  BUG
Influences：  
fix: remove redundant detection after reload
#3298


wenyh1 (approved):
lgtm

PR #3334, opened 2022-08-03
fix: newly added instance needs to create a connection pool immediately

LUAgam:
Reason:  
  BUG inner-1847
Type:  
  BUG
Influences：  
fix: newly added instance needs to create a connection pool immediately


PanternBao (approved):
lgtm

PR #3338, opened 2022-08-04
fix: heartbeat status is STOP_STATUS after ha disable

LUAgam:
Reason:  
  BUG inner-1847
Type:  
  BUG
Influences：  
fix: heartbeat status is STOP_STATUS after ha disable

### DIFF
--- a/src/main/java/com/actiontech/dble/backend/datasource/PhysicalDbGroup.java
+++ b/src/main/java/com/actiontech/dble/backend/datasource/PhysicalDbGroup.java
@@ -458,7 +458,7 @@ public class PhysicalDbGroup {
             if (ds.isAlive() && (!checkSlaveSynStatus() || ds.canSelectAsReadNode())) {
                 okSources.add(ds);
             } else {
-                if (!ds.isAlive()) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/actiontech/dble/backend/datasource/PhysicalDbGroup.java
```diff
@@ -458,7 +458,7 @@ private List<PhysicalDbInstance> getRWDbInstances(boolean includeWrite) {
             if (ds.isAlive() && (!checkSlaveSynStatus() || ds.canSelectAsReadNode())) {
                 okSources.add(ds);
             } else {
-                if (!ds.isAlive()) {
```
yanhuqing666, line 461: I think it also need to print log when `!ds.isAlive()`

LUAgam: too many logs printed when !ds.isAlive()
inner-1845

LUAgam:
need update: reduce the frequency of log printing

PanternBao (approved):
lgtm

PR #3340, opened 2022-08-05
[inner-1872] fix show trace bug

wenyh1:
Reason:  
  BUG inner 1872
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3341, opened 2022-08-08
fix: config is synced to ha's memory

LUAgam:
Reason:  
  BUG inner-1871&1874
Type:  
  BUG
Influences：  
fix: config is synced to ha's memory


wenyh1 (approved):
lgtm

PR #3343, opened 2022-08-08
perf: reduce the number of test connections

LUAgam:
Reason:  
  BUG inner-1881/1871
Type:  
  BUG
Influences：  
perf: reduce the number of test connections


PanternBao (approved):
lgtm

PR #3344, opened 2022-08-09
fix  mysql heartbeat fails to recover for a long time due to VIP drift inner 1495

ylinzhu:
Reason:  
   Improve 
Type:  
  Improve  
Influences：  
   fix xx


LUAgam (approved):
lgtm

PR #3345, opened 2022-08-10
[inner-1873] scene of the concurrent, FrontActiveRatioStat's error

wenyh1:
Reason:  
  BUG inner 1873
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3346, opened 2022-08-10
fix: sync is required when sharding.xml is empty

LUAgam:
Reason:  
  BUG inner-1886
Type:  
  BUG
Influences：  
fix: sync is required when sharding.xml is empty


wenyh1 (approved):
ok

PR #3348, opened 2022-08-11
[inner-1844] Fix the round/truncate function problem（cherry pick）

wenyh1:
Reason:  
  BUG similar #3330 
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3350, opened 2022-08-11
fix: the connection pool must be recycled when deleting dbInstance

LUAgam:
Reason:  
  BUG inner-1887
Type:  
  BUG
Influences：  
fix: the connection pool must be recycled when deleting dbInstance


PanternBao (commented):
lgtm

PR #3354, opened 2022-08-12
[inner-1839] strengthen judgment

wenyh1:
Reason:  
  BUG inner 1839
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #3356, opened 2022-08-15
inner-1501: fix NullPointException

PanternBao:
Reason:  
  BUG #inner-1501.  
Type:  
  BUG
Influences：  
  fix NullPointException


dcy10000 (approved):
ok

PR #3359, opened 2022-08-16
fix: config is synced to ha's memory(cherry-pick)

LUAgam:
Reason:
  BUG inner-1874， similar #3341 
Type:
  BUG
Influences：
fix: config is synced to ha's memory


wenyh1 (approved):
ok

PR #3362, opened 2022-08-16
[inner-1812] fix: when natural join could not find the column

wenyh1:
Reason:  
  BUG inner 1812
Type:  
  BUG/Improve  
Influences：  
   fix xx


wenyh1:
draft reason: there may be other issue to deal with together

wenyh1:
There are other PR

PR #3363, opened 2022-08-16
fix: use thread-safe collections(cherry-pick from 2261)

LUAgam:
Reason:  
  BUG inner-1880
Type:  
  BUG
Influences：  
fix: use thread-safe collections
#2261


wenyh1 (approved):
ok

PR #3365, opened 2022-08-16
[inner-1812&1811&1810]  fix : when natural join,  fix 'could not find the column' and 'duplicate column' and adjust columns order

wenyh1:
Reason:  
  BUG inner 1812、1811
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3366, opened 2022-08-17
fix: use lowercase for status

LUAgam:
Reason:  
  BUG #3357
Type:  
  BUG
Influences：  
fix: use lowercase for status


PanternBao (approved):
lgtm

PR #3367, opened 2022-08-17
fix: newly added instance needs to create a connection pool immediately(cherry-pick)

LUAgam:
Reason:
  BUG inner-1847、inner-1874
Type:
  BUG
Influences：
fix: newly added instance needs to create a connection pool immediately #3357
heartbeat status is STOP_STATUS after ha disable
fix: config is synced to ha's memory #3359
fix: use lowercase for status #3366


PanternBao (approved):
lgtm

PR #3375, opened 2022-08-19
inner-1900: use table alias to replace table index

PanternBao:
Reason:  
  BUG #inner 1900.
Type:  
  BUG  
Influences：  
  inner-1900: use table alias to replace table index


ylinzhu (approved):
ok

PR #3379, opened 2022-08-23
inner-1904: fix direct memory leak (#3378)

PanternBao:
Reason:  
  BUG #inner-1904.
Type:  
  BUG
Influences：  
  inner-1904: fix direct memory leak (#3378)


LUAgam (approved):
ok

PR #3380, opened 2022-08-23
[inner-1830] fix：result of the a group_concat() is incorrect

wenyh1:
Reason:  
  BUG inner 1830
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #3391, opened 2022-08-29
[inner-1841] fix: group_concat()'s sort problem

wenyh1:
Reason:  
  BUG 1841
Type:  
  BUG/Improve  
Influences：  
   fix xx


src/main/java/com/actiontech/dble/plan/common/item/function/sumfunc/ItemFuncGroupConcat.java
```diff
@@ -225,6 +227,16 @@ public boolean pushDownAdd(RowDataPacket row) {
         throw new RuntimeException("not implement");
     }
 
+    @Override
+    public void fixRefer(ReferContext context) {
```
PanternBao, line 231: try to move to fixOrders method

wenyh1: ok

PanternBao (approved):
lgtm

PR #3393, opened 2022-08-30
fix when joinStrategyType is 2 inner 1772 & 1804 & 1816 (cherry pick)

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG/Improve  
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3394, opened 2022-08-30
fix: change user protocol

LUAgam:
Reason:  
  BUG inner-1894
Type:  
  BUG
Influences：  
fix: change user protocol


LUAgam:
need to reopen

PR #3409, opened 2022-09-13
[inner-1614] add a detection switch for 'xa recover'

wenyh1:
Reason:  
  Improve inner-1614
Type:  
  Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3411, opened 2022-09-14
inner-1516:json decode function support

dcy10000:

Reason:  
    Improve :inner-1516:json decode function support
Type:  
  Improve  
Influences：  
  json


LUAgam (approved):
ok

PR #3417, opened 2022-09-21
[inner-1833] support for 'group/order by' columns's refer table size > 1

wenyh1:
Reason:  
  BUG inner 1833
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3421, opened 2022-09-27
issue 3420: dble parse date parameter error in com_execute packet

PanternBao:
Reason:  
  BUG #3420. 
Type:  
  BUG
Influences：  
  dble parse date parameter error in com_execute packet


wenyh1 (approved):
lgtm

PR #3422, opened 2022-09-28
inner-1937：fix json return 0 when unmatched && optimize json unquote

dcy10000:
Reason:  
  BUG inner-1937
Type:  
  BUG
Influences：  
  fix json return 0 when unmatched && optimize json unquote


PanternBao (approved):
lgtm

PR #3424, opened 2022-09-29
inner 1489: readonly transaction can push to slave

PanternBao:
Reason:  
    Improve #inner 1489.  
Type:  
  BUG/Improve  
Influences：  
  inner 1489: readonly transaction can push to slave


ylinzhu (approved):
ok

PR #3432, opened 2022-10-10
turn on testWhileIdle in db_template.xml

PanternBao:
Reason:  
  Improve.  
Type:  
  Improve  
Influences：  
  turn on testWhileIdle in db_template.xml


dcy10000 (approved):
ok

PR #3433, opened 2022-10-11
feat: supports logical create and drop of virtual schema

LUAgam:
Reason:  
  Improve inner-1852
Type:  
  Improve 
Influences：  
feat: supports logical create and drop of virtual schema


wenyh1 (approved):
ok

PR #3437, opened 2022-10-17
inner 1489 - push readonly transaction to slave in rw

PanternBao:
Reason:  
     Improve #inner 1489.  
Type:  
  Improve  
Influences：  
  push readonly transaction to slave


ylinzhu (approved):
ok

PR #3438, opened 2022-10-17
fix:  improve the connection pool to supplement the connection logic

LUAgam:
Reason:  
  Improve inner-1867&1933
Type:  
  Improve  
Influences：  
fix:  improve the connection pool to supplement the connection logic


PanternBao (approved):
lgtm

PR #3441, opened 2022-10-19
inner 1927 - avoid timeout heartbeat connection remaining

PanternBao:
Reason:  
  BUG #inner 1927. 
Type:  
  BUG
Influences：  
  avoid timeout heartbeat connection remaining


LUAgam (approved):
ok

PR #3445, opened 2022-10-21
RW for sharding User supports local read inner 1951

ylinzhu:
Reason:  
  Improve 
Type:  
  Improve  
Influences：


wenyh1 (approved):
lgtm

PR #3446, opened 2022-10-24
inner-1937:json fix: arr[0] return self

dcy10000:

Reason:  
  BUG inner-1937
Type:  
  BUG
Influences：  
  json fix: arr[0] return self


LUAgam (approved):
ok

PR #3447, opened 2022-10-25
fix: dble_schema display fields

LUAgam:
Reason:  
  BUG inner-1852
Type:  
  BUG
Influences：  
fix: dble_schema display fields


wenyh1 (approved):
ok

PR #3448, opened 2022-10-26
inner-1945: print error if no handler

dcy10000:

Reason:  
  BUG inner-1945
Type:  
  BUG
Influences：  
   print error if no handler


wenyh1 (approved):
ok

PR #3453, opened 2022-11-01
inner 1497 - avoid concurrency between close and xa prepare

PanternBao:
Reason:  
  BUG #inner 1497.
Type:  
  BUG
Influences：  
  avoid concurrency between close and xa prepare


ylinzhu (approved):
ok

PR #3455, opened 2022-11-01
inner 1927 - fix initial status  to quit status

PanternBao:
Reason:  
  BUG #inner 1927.
Type:  
  BUG
Influences：  
  fix initial status  to quit status


LUAgam (approved):
ok

PR #3457, opened 2022-11-03
fix: alert when xa retry stop

LUAgam:
Reason:  
  BUG inner-1423
Type:  
  BUG
Influences：  
fix: alert when xa retry stop


dcy10000 (approved):
lgtm

PR #3459, opened 2022-11-03
inner 1970 - handle multi-statement error when sql contains ddl

PanternBao:
Reason:  
  BUG #inner 1970. 
Type:  
  BUG
Influences：  
  handle multi-statement error when sql contains ddl


wenyh1 (approved):
ok

wenyh1 (approved):
ok

PR #3462, opened 2022-11-07
fix memory leak inner 1860

ylinzhu:
fix npe when connection concurrency is closed

Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #3467, opened 2022-11-08
fix: strategies to increase slow query blocking

LUAgam:
Reason:
  Improve inner-1763
Type:
  Improve
Influences：
pref: strategies to increase slow query blocking


wenyh1 (approved):
ok

PR #3470, opened 2022-11-09
fix npe when the frontend connection is closed, the statement is still send

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：


LUAgam (approved):
ok

PR #3471, opened 2022-11-09
fix npe  when innerRelease and close are concurrent inner 1883

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3472, opened 2022-11-10
inner-1898: release switch & inner-1897:prevent sync release hang 

dcy10000:
Reason:  
  BUG inner-1898 && inner-1897
Type:  
  BUG/Improve  
Influences：  
   release switch


ylinzhu (approved):
ok

PR #3479, opened 2022-11-15
inner-1898: release switch & inner-1897:prevent sync release hang

ylinzhu:
fix

Reason:  
  Improve 
Type:  
  /Improve  
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3480, opened 2022-11-15
[inner-1993] in rwsplit, filter 1064 syntax errors

wenyh1:
Reason:  
  BUG inner 1993
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3481, opened 2022-11-15
fix: use the state value before restoration to judge

LUAgam:
Reason:  
  BUG inner-1933
Type:  
  BUG
Influences：  
fix: use the state value before restoration to judge


PanternBao (approved):
lgtm

PR #3482, opened 2022-11-15
fix: exposed errors during dryrun

LUAgam:
Reason:  
  BUG inner-1976/inner-1995
Type:  
  BUG
Influences：  
fix: exposed errors during dryrun
dble_variables display queryForUpdateMaxRowsSize field


dcy10000 (approved):
ok

PR #3489, opened 2022-11-21
fix  performance optimization inner 1922

ylinzhu:
Reason:  
  Improve 
Type:  
  Improve  
Influences：  
   fix xx


wenyh1 (approved):
lgtm

PR #3490, opened 2022-11-21
fix: explain result is wrong

LUAgam:
Reason:  
  BUG inner-1995&inner-1997
Type:  
  BUG
Influences：  
fix: explain result is wrong
check the validity of queryForUpdateMaxRowsSize value


wenyh1 (approved):
lgtm

PR #3493, opened 2022-11-23
[inner-1783] adjust transaction's debug log

wenyh1:
Reason:  
  BUG inner 1783
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3494, opened 2022-11-24
inner 1959 - avoid the case that connectionClose thread is updating flag but other thread had finished

PanternBao:
Reason:  
  BUG #inner 1959.
Type:  
  BUG
Influences：


wenyh1 (approved):
lgtm

PR #3497, opened 2022-11-25
inner 2022 - promise that add unResponseRrns is atomic

PanternBao:
Reason:  
  BUG #inner 2022.
Type:  
  BUG
Influences：  
  promise that add unResponseRrns is atomic


LUAgam (approved):
ok

PR #3500, opened 2022-11-29
fix: join on converts to where syntax

LUAgam:
Reason:  
  BUG inner-1995/inner-2024/inner-2027
Type:  
  BUG
Influences：  
fix: join on converts to where syntax
explain removes extra spaces


dcy10000 (approved):
ok

PR #3501, opened 2022-11-29
fix: type conversion error

LUAgam:
Reason:  
  BUG inner-2025
Type:  
  BUG
Influences：  
fix: type conversion error


wenyh1 (approved):
ok

PR #3502, opened 2022-11-29
Inner 2016&1987 fix bug for memory-monitor

dcy10000:
Reason:  
  BUG Inner 2016&1987 
Type:  
  BUG
Influences：  
  fix bug for memory-monitor


wenyh1 (approved):
ok

PR #3504, opened 2022-11-29
inner-2020: prevent repeat json function call.

dcy10000:

Reason:  
  BUG  inner-2020
Type:  
  BUG
Influences：  
  prevent repeat json function call.


ylinzhu (approved):
ok

PR #3507, opened 2022-11-30
fix: add the determination of whether ssl is enabled（cherry-pick）

LUAgam:
Reason:
  BUG inner-2028
Type:
  BUG
Influences：
fix: add the determination of whether ssl is enabled
when the buffer is always full but no new data has been read, an error needs to be reported
#3505


wenyh1 (approved):
ok

PR #3508, opened 2022-11-30
fix: type conversion error(cherry-pick)

LUAgam:
Reason:
  BUG inner-2025
Type:
  BUG
Influences：
fix: type conversion error
#3501


wenyh1 (approved):
ok

PR #3510, opened 2022-12-01
[cherry-pick]inner-1956:fix MultiNodeSelectHandler hang &inner-1945: print error if no handler 

dcy10000:
Reason:  
  BUG [cherry-pick]inner-1956:fix MultiNodeSelectHandler hang &inner-1945: print error if no handler 
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3512, opened 2022-12-01
inner-2032: fix memory leak when kill query

dcy10000:

Reason:  
  BUG inner-2032
Type:  
  BUG/Improve  
Influences：  
  fix memory leak when kill query


ylinzhu (approved):
ok

PR #3513, opened 2022-12-01
fix: group by to remove duplicate fields

LUAgam:
Reason:  
  BUG inner-2031/inner-2024
Type:  
  BUG
Influences：  
fix: group by to remove duplicate fields
Subquery fields are not allowed to be updated


ylinzhu (approved):
ok

PR #3514, opened 2022-12-02
fix:try splitting and sending when transparent transmission is not supported

LUAgam:
Reason:  
  BUG inner-2027
Type:  
  BUG
Influences：  
fix:try splitting and sending when transparent transmission is not supported


ylinzhu (approved):
ok

PR #3516, opened 2022-12-02
fix: select another dbInstance to try again when the dbInstance is unavailable

LUAgam:
Reason:  
  BUG inner-2003
Type:  
  BUG
Influences：  
fix: select another dbInstance to try again when the dbInstance is unavailable


dcy10000 (approved):
lgtm

PR #3518, opened 2022-12-05
inner 1959 - fix remaining bug

PanternBao:
Reason:  
  BUG #inner 1959.
Type:  
  BUG
Influences：  
  last pr leads to the bug


wenyh1 (approved):
ok

PR #3525, opened 2022-12-07
inner 2040 - fix refactor code bug

PanternBao:
Reason:  
  BUG #inner 2040. 
Type:  
  BUG
Influences：  
  fix remaining bug


LUAgam (approved):
ok

PR #3527, opened 2022-12-07
inner-1956-bugfix: fix for back-rw block

dcy10000:

Reason:  
  BUG inner-1956-bugfix: fix for back-rw block
Type:  
  BUG 
Influences：  
   fix xx


PanternBao (approved):
ok

PR #3533, opened 2022-12-08
[inner-1929] adjust reload log

wenyh1:
Reason:  
  BUG inner  1929
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #3535, opened 2022-12-09
fix: Set the jdk version of codeQL

LUAgam:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
fix: Set the jdk version of codeQL


dcy10000 (approved):
ok

PR #3537, opened 2022-12-09
inner-2004: fix for max_allowed_packet in multi select

dcy10000:

Reason:  
  BUG  inner-2004x
Type:  
  BUG
Influences：  
    fix for max_allowed_packet in multi select


ylinzhu (approved):
ok

PR #3539, opened 2022-12-09
inner 1959 - avoid the case that connectionClose thread is updating flag but other thread had finished

PanternBao:
Reason:  
  BUG #inner 1959.  
Type:  
  BUG
Influences：  
  avoid the case that connectionClose thread is updating flag but other thread had finished


PanternBao:
can't need fix privious version

PR #3541, opened 2022-12-10
fix: front-end determines that the packet size cannot exceed the set value(cherry-pick)

LUAgam:
Reason:
  BUG inner-2004
Type:
  BUG
Influences：
fix: front-end determines that the packet size cannot exceed the set value


ylinzhu (approved):
ok

PR #3544, opened 2022-12-13
fix: not display note-level warnings

LUAgam:
Reason:  
  Improve    
Type:  
  Improve  
Influences：  
fix: not display note-level warnings


PanternBao (approved):
ok

PR #3565, opened 2022-12-22
inner 1959 - fix bug about lock table

PanternBao:
Reason:  
  BUG #inner 1959. 
Type:  
  BUG
Influences：  
  fix bug about lock table


wenyh1 (approved):
lgtm

PR #3567, opened 2022-12-23
database is not checked during login inner 2060

ylinzhu:
Reason:  
  Improve  
Type:  
  Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3568, opened 2022-12-23
fix: hint statement use executeSql（cherry-pick）

LUAgam:
Reason:
  BUG  inner-2046 inner-902/906  inner-603/605/606
Type:
  BUG
Influences：
fix: hint statement use executeSql
fix: wrong judgment when parsing reload


PanternBao (approved):
ok

PR #3570, opened 2022-12-27
Inner 711&631: fixed:multi-query cause timeout and 'CALL' query causes hang(cherry-pick)

LUAgam:
Reason:
  In RwSplit scenarios, when using JDBC with URL param allowMultiQueries, send multi-query cause timeout.Error reports:(2013, 'Lost connection to MySQL server during query')
Type:
  BUG inner-711
Influences：
   fixed.

Reason:
  In RwSplit scenarios, the CALL query causes hang.
Type:
  BUG inner-631
Influences：
   fixed.


dcy10000 (approved):
ok

PR #3582, opened 2023-01-06
fix: runnableMap does not store complexQueryWorker threads

LUAgam:
Reason:  
  BUG inner-2072
Type:  
  BUG
Influences：  
fix: runnableMap does not store complexQueryWorker threads


dcy10000 (approved):
lgtm

PR #3583, opened 2023-01-06
fix: runnableMap does not store complexQueryWorker threads(cherry-pick)

LUAgam:
Reason:
  BUG inner-2072
Type:
  BUG
Influences：
fix: runnableMap does not store complexQueryWorker threads
#3582


wenyh1 (approved):
ok

PR #3593, opened 2023-01-12
move root user blacklist to another user

PanternBao:
Reason:  
   Improve #3590 .  
Type:  
  Improve  
Influences：  
  move root user blacklist to another user


LUAgam (approved):
ok

PR #3600, opened 2023-01-30
fix npe when the frontend connection is closed, the statement is still send

ylinzhu:
https://github.com/actiontech/dble/pull/3470
Reason:  
  BUG
Type:  
  BUG 
Influences：  
  inner 1879


wenyh1 (commented):
ok

PR #3601, opened 2023-02-01
fix: grpc timeout configuration

LUAgam:
Reason:  
  BUG inner-2089
Type:  
  BUG
Influences：  
fix: grpc timeout configuration


PanternBao (approved):
lgtm

PR #3602, opened 2023-02-02
fix manage execute sql hang inner 2091

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：


ylinzhu:
Wait for verification

ylinzhu:
This fix is the same as https://github.com/actiontech/dble/pull/3612, and https://github.com/actiontech/dble/pull/3612  has been validated, so close the commit

PR #3604, opened 2023-02-03
[inner-2092]: rwSplit's big package problem

wenyh1:
Reason:  
  BUG inner-2092
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (approved):
Lgtm

PR #3606, opened 2023-02-07
[inner-2096] fix: the problem of hung in the 16M's big package

wenyh1:
Reason:  
  BUG inner 2096 ,similar #3605 
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (approved):
ok

PR #3608, opened 2023-02-08
fix timer thread  hang inner 1790 & 2037 (#3287)

ylinzhu:
Conflicts:
	src/main/java/com/actiontech/dble/backend/mysql/nio/handler/NewConnectionRespHandler.java
Reason:  
  BUG 
Type:  
  BUG
Influences：


LUAgam (approved):
ok

PR #3612, opened 2023-02-09
fix: close the connection that is being created and the creation time exceeds the connectionTimeout value

LUAgam:
Reason:  
  BUG inner-1716&inner-1955&inner-2091
Type:  
  BUG
Influences：  
fix: close the connection that is being created and the creation time exceeds the connectionTimeout value


PanternBao (approved):
lgtm

PR #3620, opened 2023-02-20
[inner-2095] Log4j upgrade to 2.18.0 version (#3605)

wenyh1:
Reason:  
  BUG inner 2095 ,similar #3605 
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3627, opened 2023-02-27
inner-2114: modify error msg from no handle to drop

dcy10000:
Reason:  
  Improve inner-2114: modify error msg from no handle to drop
Type:  
  Improve
Influences：  
   log


ylinzhu (approved):
ok

PR #3629, opened 2023-03-08
fix: adjust the identification process of the ssl protocol package

LUAgam:
Reason:  
  BUG inner-2029
Type:  
  BUG
Influences：  
+ fix: adjust the identification process of the ssl protocol package 
+ solve the memory leak problem


wenyh1 (approved):
lgtm

PR #3630, opened 2023-03-10
inner-2140: fix the multi-thread bug for xa

dcy10000:
Reason:  
  BUG inner-2140
Type:  
  BUG
Influences：  
  fix the multi-thread bug for xa


ylinzhu (approved):
ok

PR #3631, opened 2023-03-14
inner-2142:fix npe for inner-2140

dcy10000:
Reason:  
  BUG inner-2142
Type:  
  BUG
Influences：  
   :fix npe for inner-2140


ylinzhu (approved):
ok

PR #3632, opened 2023-03-15
[inner-2144] fix：sql_log's view with conditional query error

wenyh1:
Reason:  
  BUG inner 2144
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3633, opened 2023-03-16
fix: when checking the back_log, dbInstance does not correspond to KeyVariables

LUAgam:
Reason:  
  BUG inner-2149
Type:  
  BUG
Influences：  
fix: when checking the back_log, dbInstance does not correspond to KeyVariables


ylinzhu (approved):
ok

PR #3639, opened 2023-03-23
inner-2158:add connection create log info

dcy10000:
Reason:  
  Improve  inner-2158
Type:  
  Improve
Influences：  
   add connection create log info


LUAgam (approved):
ok

PR #3641, opened 2023-03-24
[inner-2157] fix lost connection bug from ddl

wenyh1:
Reason:  
  BUG inner 2157
Type:  
  BUG/Improve  
Influences：  
   fix lost connection bug from ddl


ylinzhu (approved):
ok

PR #3645, opened 2023-03-29
[inner-1576&2161] adjust the trigger global sequence's load

wenyh1:
Reason:  
  BUG inner 1576、2161
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
lgtm

PR #3650, opened 2023-04-03
fix: variable display and default settings

LUAgam:
Reason:  
  BUG inner-2089&inner-2051
Type:  
  BUG
Influences：  
fix: variable display and default settings


dcy10000 (approved):
ok

PR #3659, opened 2023-04-13
fixed 'show @@connection.sql.status & show @@connection_pool' in python report ValueError inner 2188 

ylinzhu:
Reason:  
  BUG
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3569, opened 2022-12-26
issue 3238 - fix cpu 100%

PanternBao:
Reason:  
  BUG #3238 .
Type:  
  BUG
Influences：  
  fix cpu 100%

Issue #1687, opened 2020-03-06
optimize a certain sql trace time for route calculate too long 


irene-coming:
* **dble version：**  
5.6.29-dble-9.9.9.9-26da2a0733b3f17a54aa2a3a7a747b6e6c966449-20200304091656
* **preconditions ：**  
sharding_2_t1 is a two sharding table
* **configs：** 

**schema.xml**  

```


```

**rule.xml**  

```


```

**server.xml**  

```


```

* **steps：**    
    step1.  execute query in dble:
```mysql
create table sharding_2_t1(id int,
 normal_col_1 varchar(30),
 normal_col_2 varchar(30),
 normal_col_3 varchar(30),
 normal_col_4 varchar(30),
 normal_col_5 varchar(30),
 normal_col_6 varchar(30),
 normal_col_7 varchar(30),
 normal_col_8 varchar(30),
 normal_col_9 varchar(30),
 normal_col_10 varchar(30),
 normal_col_11 varchar(30),
 normal_col_12 varchar(30),
 normal_col_13 varchar(30),
 normal_col_14 varchar(30),
 normal_col_15 varchar(30),
 min_val1   varchar(30),
  min_val2   varchar(30),
   min_val3   varchar(30),
    min_val4   varchar(30),
     min_val5   varchar(30),
      min_val6   varchar(30),
       min_val7   varchar(30),
  max_val1   varchar(30),
  max_val2   varchar(30),
   max_val3   varchar(30),
    max_val4   varchar(30),
     max_val5   varchar(30),
      max_val6   varchar(30),
       max_val7   varchar(30)
 );
 
  SET TRACE=1;
  
  select *
 from sharding_2_t1
 where id = '12'
   and normal_col_1 = '0000345171'
   and normal_col_2 = '1'
   and normal_col_3 = '100'
   and normal_col_4 = '1'
   and normal_col_5 = '001'
   and normal_col_6 = ''
   and normal_col_7 = ''
   and normal_col_8 = ''
   and normal_col_9 = 1
   and normal_col_10 = 1
   and normal_col_11 = 1
   and normal_col_12 = 1
   and normal_col_13 = 0
   and normal_col_14 = 0
   and normal_col_15 = 0
   and ((normal_col_9 = 0
     or (normal_col_9 = 1
       and min_val1 <= 1
       and max_val1 >= 1)
     or (normal_col_9 = 2
       and min_val1 < 1
       and max_val1 > 1)
     or (normal_col_9 = 3
[...]

Issue #1542, opened 2019-12-11
need compatible with all mysql versions for system variable "transaction_isolation(>=5.7.20)" and "tx_isolation(<5.7.20)" 


wjl1619:
* **dble version：**  
* **preconditions ：**  
no
* **configs：** 

**schema.xml**  

```


```

**rule.xml**  

```


```

**server.xml**  

```


```

* **steps：**    
    step1.  
* **expect result：**  
    1.
* **real result：**  
    1.
* **supplements：**  
    1.

yanhuqing666:
delay

yanhuqing666:
DBLE WILL accept  "transaction_isolation(>=5.7.20)" and "tx_isolation(<5.7.20)" and "transaction_read_only(>=5.7.20)" and "tx__read_only(<5.7.20)".
AND use `tx` to send to MySQL(<8.0 ) AND use  `transaction` to send to MySQL(>=8.0 )

PR #3665, opened 2023-04-20
2.19.03/lts

TommyZC:
Reason:  
  issue 3238 - fix cpu 100% (#3569)
        Modify the culling logic of show @@sql.high, let it use time dimension firstly
        add TOTAL_TIME parameter into "show @@sql.high"
        Close sqlRwStat, sqlLastStatSQL, sqlLargeStat, big size sql functions and only leave sqlHighStat function when useSqlStat=1
        #1542 support 8.0 transaction_ variables
        #1687 simple route result (#1841)
        #1687 change or
        #1687 remove where unit
        #1687 some bugs
        #1687 alias without BackQuote
        #1687 two table always false
        #1687 rename field in
        remove useless String
        add "digest" parameter of SQL's hashcode into "show @@sql.high"
Type:  
  BUG and Improve


LUAgam (approved):
ok

PR #3669, opened 2023-04-23
inner-2196: UPDATE SQL support sharding join global

dcy10000:
Reason:  
  BUG inner-2196
Type:  
  BUG
Influences：  
  UPDATE SQL support sharding join global


PanternBao (approved):
lgtm

PR #3670, opened 2023-04-26
fix: the timing of obtaining the number of waiting threads in the connection pool is wrong

LUAgam:
Reason:  
  BUG inner-1867/inner-1740
Type:  
  BUG
Influences：  
fix: the timing of obtaining the number of waiting threads in the connection pool is wrong


PanternBao (approved):
lgtm

PR #3678, opened 2023-05-04
[inner-2208] display column aliases correctly

wenyh1:
Reason:  
  BUG inner 2208
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3679, opened 2023-05-04
fix: service is set to null and connection close is not atomic

LUAgam:
Reason:  
  BUG inner-1740
Type:  
  BUG
Influences：  
fix: service is set to null and connection close is not atomic


PanternBao (approved):
lgtm

PR #3683, opened 2023-05-08
revert the merge of actiontech#1687 because it sloves the problem of …

TommyZC:
…slow parsing of SQL statements, but causes the new problems, for example, some SQL statements like "union select" can't be nomally parsed.

Reason:  
  BUG  
Type:  
  BUG
Influences：  
  merge代码后解决了解析SQL缓慢的问题，但使用自动化测试工具进行测试，发现对于一些复杂语句，例如union select语句，无法正常解析和执行。


ylinzhu (approved):
ok

PR #3675, opened 2023-04-27
[inner-2205] should update the latest sequence_conf.properties when sequenceHandlerType=4 and executing 'reload @@config'

wenyh1:
…

Reason:  
  BUG inner 2205
Type:  
  BUG/Improve  
Influences：  
   fix xx

PR #3684, opened 2023-05-08
2161&2205/cherry pick 22.01

wenyh1:
Reason:  
  similar #3645  #3675 
Type:  
  BUG/Improve  
Influences：  
   fix xx


wenyh1:
error

PR #3686, opened 2023-05-09
fix jdbc compatible pre-delivery statements error inner 2212

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


PanternBao (approved):
lgtm

PR #3689, opened 2023-05-15
fix timer thread  hang inner 1790 & 2037 (#3287)

ylinzhu:
cherrk pick from  https://github.com/actiontech/dble/pull/3287
Reason:  
  BUG
Type:  
  BUG
Influences：  
   fix xx


dcy10000 (approved):
lgtm

PR #3692, opened 2023-05-15
fix: peel version check module 

LUAgam:
Reason:  
  BUG inner-2221
Type:  
  BUG
Influences：  
fix: peel version check module 
when modifying the rwSplitUser-dbGroup, it is also necessary to check the version


wenyh1 (approved):
lgtm

PR #3694, opened 2023-05-17
fix: add rwSplitUser also needs to check the version

LUAgam:
Reason:  
  BUG inner-2221/inner-2226
Type:  
  BUG
Influences：  
fix: add rwSplitUser also needs to check the version 
adjust heartbeat error message


wenyh1 (approved):
lgtm

PR #3697, opened 2023-05-18
inner-2153:fix bug for  "response tabular repose error" packet

dcy10000:
Reason:  
  BUG inner-2153
Type:  
  BUG
Influences：  
  fix bug for  "response tabular repose error" packet


PanternBao (approved):
lgtm

PR #3698, opened 2023-05-18
[inner-2227] prevents 'connection-pool-evictor-thread' thread from exiting unexpectedly

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3702, opened 2023-05-22
[inner-2222] clear the 'sendData' resource in the transaction correctly

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3706, opened 2023-05-24
fix: utf8mb4_0900_ai_c is only supported when the dble version is greater than or equal to 8.0.3

LUAgam:
Reason:  
  BUG inner-2051
Type:  
  BUG
Influences：  
fix: utf8mb4_0900_ai_c is only supported when the dble version is greater than or equal to 8.0.3


PanternBao (approved):
lgtm

PR #3708, opened 2023-05-26
fix: ssl not wrap all the content in the buffer

LUAgam:
Reason:  
  BUG inner-2235
Type:  
  BUG
Influences：  
fix: ssl not wrap all the information in the buffer


PanternBao (approved):
lgtm

PR #3710, opened 2023-05-26
fix: rw-split user are not limited by maxPacketSize of dble

LUAgam:
Reason:  
  BUG inner-2048
Type:  
  BUG
Influences：  
fix: rw-split user are not limited by maxPacketSize of dble


wenyh1 (approved):
ok

PR #3711, opened 2023-05-29
[inner-2215] in execute protocol, prepare sql need to be recorded in sql_log

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3716, opened 2023-05-31
fix parameter check inner 2239

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


src/main/java/com/actiontech/dble/config/model/SystemConfig.java
```diff
@@ -1903,8 +1904,8 @@ public int getTcpKeepIdle() {
         return tcpKeepIdle;
     }
 
-    public void setTcpKeepIdle(int tcpKeepIdle) {
-        if (tcpKeepIdle > 0) {
+    public void setTcpKeepIdle(int tcpKeepIdle) throws IOException {
+        if (tcpKeepIdle > 0 && DbleSocketOptions.check(DbleSocketOptions.TCP_KEEP_IDLE, tcpKeepIdle)) {
```
wenyh1, line 1908: Advice: complex check should not be performed in the reflection method

wenyh1 (approved):
ok

PR #3717, opened 2023-06-01
fix: after the slow log queue is full, discard the subsequent records that need to be added(cherry-pick)

LUAgam:
Reason:
  BUG inner-2227/inner-2230
Type:
  BUG
Influences：
fix: after the slow log queue is full, discard the subsequent records that need to be added
after abnormal disk flush, fileChannel will not close
#3704


ylinzhu (approved):
ok

PR #3685, opened 2023-05-08
2161&2205/cherry pick 22.01

wenyh1:
Reason:  
  BUG https://github.com/actiontech/dble/pull/3645 https://github.com/actiontech/dble/pull/3675
Type:  
  BUG/Improve  
Influences：  
   fix xx

PR #3718, opened 2023-06-01
[inner-2205] should update the latest sequence_conf.properties when sequenceHandlerType=4 and executing 'reload @@config' (cherry pick from #3685)

wenyh1:
Reason:  
  BUG similar #3675 
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3721, opened 2023-06-02
fix: after abnormal disk flush, fileChannel will not close(cherry-pick) 

LUAgam:
Reason:
  BUG inner-2230
Type:
  BUG
Influences：
after abnormal disk flush, fileChannel will not close
#3704


ylinzhu (approved):
ok

PR #3730, opened 2023-06-09
fix: DMP-5667 and no cluster ha with reload

LUAgam:
fix: cherry-pick
DMP-5667 (#2181) 
no cluster ha with reload (#2182)


wenyh1 (approved):
ok

PR #3733, opened 2023-06-13
[inner-2161] sequence configuration should not be instantiated when dryrun (cherry pick from #3685)

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (approved):
ok

PR #3735, opened 2023-06-15
[inner-2259] fix: large and small table join scenario, 'show trace' result is empty

wenyh1:
Reason:  
  BUG inner-2259
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (approved):
lgtm

PR #3740, opened 2023-06-16
btrace - add slowlog entry

PanternBao:
Reason:  
  Improve.  
Type:  
  Improve  
Influences：  
  btrace - add slowlog entry


ylinzhu (approved):
ok

PR #3754, opened 2023-06-27
[inner-2111] independent management thread

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3755, opened 2023-06-27
fix: narrow down the scope of syncing dbGroup state

LUAgam:
Reason:  
  BUG inner-2256
Type:  
  BUG
Influences：  
fix: narrow down the scope of syncing dbGroup state


PanternBao (commented):
lgtm

PR #3758, opened 2023-07-03
support 0/1/false/true compatibility with or without enabled inner 2117

ylinzhu:
Reason:  
  Improve 
Type:  
      Improve  
Influences：  
   fix xx


LUAgam (approved):
lgtm

PR #3761, opened 2023-07-06
[inner-2218] during startup/dryrun/reload, apNode and associated configuration need to be detected

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


src/main/java/com/actiontech/dble/services/manager/response/DryRun.java
```diff
@@ -357,9 +399,12 @@ private static void userCheck(List<ErrorInfo> list, ServerConfig serverConfig) {
             for (UserConfig user : userMap.values()) {
                 if (user instanceof ManagerUserConfig) {
                     hasManagerUser = true;
-                } else if (user instanceof ShardingUserConfig) { // contains HybridTAUserConfig
+                } else if (user instanceof ShardingUserConfig) {
                     hasShardingUser = true;
                     schema.addAll(((ShardingUserConfig) user).getSchemas());
+                } else if (user instanceof HybridTAUserConfig) {
+                    hasShardingUser = true;
```
LUAgam, line 406: can it be reused?

LUAgam (changes requested):

PR #3763, opened 2023-07-07
fix: rows that do not need to be changed are not correctly removed during update or delete

LUAgam:
Reason:  
  BUG inner-2283
Type:  
  BUG
Influences：  
fix: rows that do not need to be changed are not correctly removed during update or delete


wenyh1 (approved):
lgtm

PR #3764, opened 2023-07-10
fix: add null value judgment

LUAgam:
Reason:  
  BUG inner-2283
Type:  
  BUG
Influences：  
fix: add null value judgment


ylinzhu (approved):
ok

PR #3765, opened 2023-07-10
fix[inner-2283]: heartbeat_keep_alive add null value judgment

LUAgam:
Reason:  
  BUG inner-2283
Type:  
  BUG
Influences：  
fix[inner-2283]: heartbeat_keep_alive add null value judgment


LUAgam:
need update

wenyh1 (approved):
ok

PR #3772, opened 2023-07-20
[inner-2211]fix: execute time-consuming tasks asynchronously

LUAgam:
Reason:  
  BUG inner-2211
Type:  
  BUG
Influences：  
fix: execute time-consuming tasks asynchronously


wenyh1 (approved):
ok

PR #3773, opened 2023-07-20
fix sql execution result error inner 590

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG/Improve  
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3775, opened 2023-07-21
[inner-2281&2301&2302] fix: when reloading, the consistency of system parameters between comparison instances is not comprehensive

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx

PR #3776, opened 2023-07-25
[inner-2281&2301&2302] fix: when reloading, the consistency of system parameters between comparison instances is not comprehensive (cherry pick from #3775)

wenyh1:
Reason:  
  inner-2281&2301&2302  similar #3775 
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3777, opened 2023-07-27
fix: determine whether an aggregate function is included in where

LUAgam:
Reason:  
  BUG inner-2307
Type:  
  BUG
Influences：  
fix: determine whether an aggregate function is included in where


wenyh1 (approved):
lgtm

PR #3780, opened 2023-07-28
[inner-2301] fix: all mysql instances are disabled, clickhouse instances are not disabled, restart and reload should not report an error(cherr pick from #3775)

wenyh1:
Reason:  
  BUG inner 2301, slimlar #3775 
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3781, opened 2023-07-28
[inner-2123] prevent threads from exiting unexpectedly

wenyh1:
Reason:  
  BUG inner 2123
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (approved):
ok

PR #3782, opened 2023-07-31
[inner-2308] fix: determine whether to deliver to AP or TP according to the rewritten SQL type

LUAgam:
Reason:  
  BUG inner-2308
Type:  
  BUG
Influences：  
fix: determine whether to deliver to AP or TP according to the rewritten SQL type


dcy10000 (approved):
ok

PR #3786, opened 2023-08-02

LUAgam:
Reason:  
  BUG inner-2311
Type:  
  BUG
Influences：  
fix: precision is missing when datetime/timestamp/time are converted to strings


ylinzhu (approved):
ok

PR #3787, opened 2023-08-02
[inner-2303] fix: not checking the version number of clickhouse and not synchronization variables to clickhouse

LUAgam:
Reason:  
  BUG inner-2303
Type:  
  BUG
Influences：  
fix: not checking the version number of clickhouse and not synchronization variables to clickhouse


wenyh1 (approved):
ok

PR #3788, opened 2023-08-03
[inner-2315] fix: handle clickhouse incompatibilities

LUAgam:
Reason:  
  BUG inner-2315
Type:  
  BUG
Influences：  
The fields returned in the clickhouse-mysql protocol do not contain information about database/table, etc.


ylinzhu (approved):
ok

PR #3790, opened 2023-08-04
[inner-2318] the lowercase on the mysql side changes, after 'reload @@config_all', the heartbeat of the dbInstance should be in the error state

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   1、fix:the lowercase on the mysql side changes, after 'reload @@config_all', the heartbeat of the dbInstance should be in the error state
        2、Unified lowercase variables


ylinzhu (approved):
ok

PR #3794, opened 2023-08-11
[inner-2320] fix: The direct route can only be selected when the shardnodes of all tables in SQL have intersection

LUAgam:
Reason:  
  BUG inner-2320
Type:  
  BUG
Influences：  
The direct route can only be selected when the shardnodes of all tables in SQL have intersection


LUAgam:
need update

PR #3795, opened 2023-08-14
[inner-2306] fix: databaseType should not be empty

LUAgam:
Reason:  
  BUG inner-2306
Type:  
  BUG
Influences：  
databaseType should not be empty
add flowHighLevel/flowLowLevel parameter value check


wenyh1 (approved):
ok

PR #3798, opened 2023-08-21
[inner-2320] fix: shardingTable can direct route only when the rules and nodes are the same

LUAgam:
Reason:  
  BUG inner-2320
Type:  
  BUG
Influences：  
shardingTable can direct route only when the rules and nodes are the same 
globalTable nodes are the same, they can be direct route


dcy10000 (approved):
lgtm

PR #3802, opened 2023-08-24
[inner-2320] fix: remove the route of noSharding

LUAgam:
Reason:  
  BUG inner-2320
Type:  
  BUG
Influences：  
remove the route of noSharding


dcy10000 (approved):
ok

PR #3804, opened 2023-08-25
[inner-2141] fix: childTable adds specifyCharset configuration item

LUAgam:
Reason:  
  BUG inner-2141
Type:  
  BUG
Influences：  
childTable adds specifyCharset configuration item


wenyh1 (approved):
ok

PR #3807, opened 2023-08-31
fix: close the connection that is being created and the creation time exceeds the connectionTimeout value

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG


LUAgam (approved):
lgtm

PR #3810, opened 2023-09-01
inner-2330&2329: release connection at once for hint in rw-transaction

dcy10000:
Reason:  
  BUG inner-2330&2329
Type:  
  BUG
Influences：  

    release connection at once for hint in rw-transaction
    fix hint in  rw-transaction


wenyh1 (approved):
lgtm

PR #3813, opened 2023-09-05
[inner-1752] fix: replace Chinese commas

LUAgam:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
 fix: replace Chinese commas


dcy10000 (approved):
ok

PR #3815, opened 2023-09-07
[inner-2339] fix: create connection timeout check should exclude connections in use (changeUser case)

LUAgam:
Reason:  
  BUG inner-2339
Type:  
  BUG
Influences：  
create connection timeout check should exclude connections in use (changeUser case)


dcy10000 (approved):
ok

PR #3817, opened 2023-09-08
[inner-2351] print thread stackTrace

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx

### DIFF
--- a/src/main/java/com/actiontech/dble/util/ExecutorUtil.java
+++ b/src/main/java/com/actiontech/dble/util/ExecutorUtil.java
@@ -61,12 +61,12 @@ public final class ExecutorUtil {
      * When the size number of threads takes a long time to execute or hangs, and the queue is full,
      * create a new thread to continue working, and the number of newly created threads does not exceed maxSize
      * <p>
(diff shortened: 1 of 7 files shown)


dcy10000 (dismissed):
ok

src/main/java/com/actiontech/dble/util/ExecutorUtil.java
```diff
@@ -66,7 +66,7 @@ private static NameableExecutor createCached(String name, int size, boolean isDa
      */
     public static NameableExecutor createCached(String name, int size, int maxSize, ThreadChecker checker) {
```
dcy10000, line 67: this api is only made for timeExecuter. 
should use a special function name to distinguish.

dcy10000 (changes requested):

dcy10000 (approved):
ok

src/main/java/com/actiontech/dble/util/ExecutorUtil.java
wenyh1: get

PR #3819, opened 2023-09-12
[inner-2060] fix: database is not checked during login(cherry-pick)

LUAgam:
Reason:  
  BUG inner-2060
Type:  
  BUG
Influences：  
database is not checked during login
#3567


ylinzhu (approved):
ok

PR #3820, opened 2023-09-12
[inner-2339] fix: create connection timeout check should exclude connections in use (cherry-pick)

LUAgam:
Reason:
  BUG inner-2339
Type:
  BUG
Influences：
create connection timeout check should exclude connections in use (changeUser case)
#3815


wenyh1 (approved):
ok

PR #3822, opened 2023-09-14
[inner-2358] manager side uses managerFrontHandlerQueue when authenticating

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (approved):
ok

PR #3826, opened 2023-10-08
[inner-2368] fix: when some nodes have been closed during the ddl execution phase, the front side needs to respond with an error.

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


LUAgam (approved):
ok

PR #3829, opened 2023-10-23
[inner-2129] When getKeyVariables, npe occasionally appears in dble.log

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx

### DIFF
--- a/src/main/java/com/actiontech/dble/config/helper/GetAndSyncDbInstanceKeyVariables.java
+++ b/src/main/java/com/actiontech/dble/config/helper/GetAndSyncDbInstanceKeyVariables.java
@@ -69,9 +69,11 @@ public class GetAndSyncDbInstanceKeyVariables implements Callable<KeyVariables> 
             while (!isFinish) {
                 boolean await = finishCond.await(ds.getConfig().getPoolConfig().getHeartbeatPeriodMillis(), TimeUnit.MILLISECONDS);
                 if (!await) {
-                    fail = true;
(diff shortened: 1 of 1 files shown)


src/main/java/com/actiontech/dble/config/helper/GetAndSyncDbInstanceKeyVariables.java
```diff
@@ -86,89 +88,85 @@ public KeyVariables call() {
...
-                        break;
-                    case "REPEATABLE-READ":
-                        keyVariables.setIsolation(Isolations.REPEATABLE_READ);
-                        break;
-                    case "SERIALIZABLE":
-                        keyVariables.setIsolation(Isolations.SERIALIZABLE);
-                        break;
-                    default:
-                        break;
+            boolean isSuccess = result.isSuccess();
```
dcy10000, line 91: reduce the lock range and using the temp variable will  more easily than this change

dcy10000 (changes requested):

src/main/java/com/actiontech/dble/config/helper/GetAndSyncDbInstanceKeyVariables.java
wenyh1: good idea

dcy10000 (approved):
lgtm

PR #3845, opened 2023-11-14
[inner-2371] fix: sql execution hang problem caused by the inability to respond to the front side due to unprocessed packets (2).

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (approved):
lgtm

PR #3850, opened 2023-12-01
fix an error that occurred during commit during load data inner 2191

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx


wenyh1 (approved):
ok

PR #3853, opened 2023-12-15
fix npe when enable the testWhileIdle parameter

ylinzhu:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix xx

### DIFF
--- a/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/ConnectionHeartBeatHandler.java
+++ b/src/main/java/com/actiontech/dble/backend/mysql/nio/handler/ConnectionHeartBeatHandler.java
@@ -88,9 +88,10 @@ public class ConnectionHeartBeatHandler implements ResponseHandler {
             }
             return;
         }
-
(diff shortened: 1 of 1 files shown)


src/main/java/com/actiontech/dble/backend/mysql/nio/handler/ConnectionHeartBeatHandler.java
```diff
@@ -88,9 +88,10 @@ public void okResponse(byte[] ok, AbstractService service) {
             }
             return;
         }
-
-        heartbeatTimeout.cancel();
-        listener.onHeartbeatSuccess((PooledConnection) service.getConnection());
+        if (!heartbeatTimeout.isExpired()) {
```
dcy10000, line 91: maybe it still triggers bug when running concurrently

dcy10000 (changes requested):

dcy10000 (approved):
ok

PR #3855, opened 2023-12-22

ylinzhu:
pick from  https://github.com/actiontech/dble/pull/3786


wenyh1 (approved):
ok

PR #3856, opened 2023-12-24
[inner-2256] fix: narrow down the scope of syncing dbGroup state

wenyh1:
Reason:  
  similar #3755 
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3859, opened 2024-01-08
[inner-2082] fix: the commit time statistics in the slow log are incorrect

wenyh1:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3860, opened 2024-01-15
fix npe when admin command is executed inner 2274

ylinzhu:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


ylinzhu:
impact surface is small, no need to repair

PR #3871, opened 2024-04-10
inner-2383: skip test other connection on some case

dcy10000:
Reason:  
  Improve  inner-2383: skip test other connection on some case
Type:  
  Improve
Influences：  
   update dbinstance


wenyh1 (approved):
lgtm

PR #3873, opened 2024-04-26
inner-2383-supplement:support delete&&support cluster

dcy10000:
Reason:  
  BUG inner-2383-supplement:
Type:  
  BUG
Influences：  
  support delete&&support cluster


ylinzhu (approved):
ok

PR #3882, opened 2024-06-14
inner-2389: fix guava incompatible

dcy10000:
Reason:  
  BUG inner-2389
Type:  
  BUG
Influences：  
   lib incompatible


ylinzhu (approved):
ok

PR #3887, opened 2024-09-13
inner-2405:can use slave connection when rwSplitMode=0.

dcy10000:
Reason:  
  BUG inner-2405:rwSplit hang after HA
Type:  
  BUG
Influences：  
  can use slave connection when rwSplitMode=0.


ylinzhu (approved):
OK

PR #3888, opened 2024-09-13
inner-2407： need init connection pool when switch with config rwsplitMode=0

dcy10000:
Reason:  
  BUG inner-2407： rwSplit user hang after HA
Type:  
  BUG
Influences：  
    need init connection pool when switch with config rwsplitMode=0


ylinzhu (approved):
OK

PR #3895, opened 2024-12-11
inner-2413: recreate heartbeat connection when timeout

dcy10000:
Reason:  
  Improve inner-2413: recreate heartbeat connection when timeout
Type:  
  Improve  
Influences：  
   fix xx


ylinzhu (approved):
ok

PR #3897, opened 2024-12-31
inner-2413: recreate heartbeat connection when timeout

dcy10000:
Reason:  
  Improve [#.  ](inner-2413: recreate heartbeat connection when timeout)
Type:  
  Improve  
Influences：


dcy10000:
not need to fix

PR #3902, opened 2025-04-21
ATK-4339&ATK-4341 :support slow log for rw user

dcy10000:
Reason:  
  Improve ATK-4339&ATK-4341 :support slow log  
Type:  
  Improve  
Influences：  
   slow log for rw


PanternBao:
lgtm

PR #3904, opened 2025-04-23
ATK-4339&ATK-4341 -supplement:fix hint doesn't work

dcy10000:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


PanternBao:
lgtm

PR #3907, opened 2025-06-03
fix "use schema" sql record  without  reach the threshold

dcy10000:
Reason:  
  BUG fix slow query
Type:  
  BUG
Influences：  
   fix xx


dcy10000:
can review later

PR #3908, opened 2025-06-18
fix slave delay calculation method

PanternBao:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix xx


dcy10000 (approved):
LGTM

PR #3909, opened 2025-06-19
atk-4459: fix unsigned binary data overflow

dcy10000:
Reason:  
  BUG atk-4459: fix unsigned binary row overflow
Type:  
  BUG
Influences：  
   prepared statement


PanternBao:
LGTM

PR #3911, opened 2025-08-04
ATK-4502: upgrade jackson jar

dcy10000:
Reason:  
  BUG upgrade jackson jar
Type:  
  BUG
Influences：  
   fix xx


PanternBao:
LGTM

PR #3916, opened 2026-01-19
fix[ATK-4637]: add Jackson dependencies and enhance MySQL delay detection logic

LUAgam:
Reason:  
  BUG 
Type:  
  BUG
Influences：  
   fix[ATK-4637]: add Jackson dependencies and enhance MySQL delay detection logic


LUAgam:
image

PR #3917, opened 2026-03-06
ATK-4657 : shutdown the old grpc channel

tian456-dot:
Reason:  
  BUG #. or Improve #.  
Type:  
  BUG/Improve  
Influences：  
   fix grpc channel leak


tian456-dot:
@dcy10000

PR #3918, opened 2026-06-29
ATK-4711 : fix dead lock when autocommit=false

tian456-dot:
Reason:  
  BUG ATK-4711
Type:  
  BUG  
Influences：  
   fix dead lock when autocommit=false


PanternBao:
LGTM

